PR #21, opened 2019-06-28
Add publicly available image to DockerHub

kyleschlosser:
The Docker image has been built and pushed to DockerHub. This pull request updates the readme and build scripts to use the new image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
 .PHONY: build deploy
 
-build:
+build: 
 	go install ./cmd/manager
 
 build-image: 
-	operator-sdk build kabanero-operator:latest
+	operator-sdk build kabanero/kabanero-operator:latest
 
 generate:
 	operator-sdk generate k8s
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,10 +1,10 @@
 .PHONY: build deploy
 
-build:
+build: 
 	go install ./cmd/manager
 
 build-image: 
-	operator-sdk build kabanero-operator:latest
+	operator-sdk build kabanero/kabanero-operator:latest
```
kaczyns: As a developer using these build targets, it would be nice if the image name could be replaced by a variable or something.  On the other hand, running `operator-sdk build imagename` is not a hard thing to do manually.  The build-image target will work great for when we're updating the public image out on dockerhub.

I don't think there needs to be a variable right now, it would be nice to have later.

kyleschlosser: I added a varible

Issue #1, opened 2019-06-20
Add specific RBAC roles to the Kabanero Operator


kyleschlosser:
The operator uses a wildcard role assignment right now, but the specific permissions need to be accounted for instead. 

Work to be completed: 
* Update the RBAC generation annotations in the source
* Run the operator-sdk scripts to regenerate the role yaml

kyleschlosser:
Regarding the work activity mentioned above `Run the operator-sdk scripts to regenerate the role yaml`, the operator sdk does not currently do such generation.

PR #22, opened 2019-06-28
Add metadata for Operator Lifecycle Manager (OLM) to source tree

mezarin:
This PR contains the initial version of the OLM artifacts (CSV and package yaml) for the Kabanero operator.  It was tested on OKD 3.11, using `operator-registry` to create a registry which was pushed to dockerhub.  OLM was installed on OKD 3.11 and used to create a catalog from the registry, and create a subscription to kabanero-operator.  OLM installs the operator and the associated roles and service account.  We verified that the OKD console could be used to create and delete a Kabanero instance.

### DIFF
--- a/deploy/olm-catalog/kabanero-operator/0.0.1/kabanero-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/kabanero-operator/0.0.1/kabanero-operator.v0.0.1.clusterserviceversion.yaml
@@ -1,0 +1,206 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  name: kabanero-operator.v0.0.1
+  namespace: placeholder
+  annotations:
+    capabilities: Basic Install
+    categories: "Integration & Delivery"
+    certified: "false"
+    containerImage: kabanero/kabanero-operator:latest
+    description: Bringings together foundational open source technologies into a modern microservices-based framework.
+    repository: https://github.com/kabanero-io/kabanero-operator
+    support: IBM
+    alm-examples: |-
+      [
+         {
+            "apiVersion":"kabanero.io/v1alpha1",
+            "kind":"Kabanero",
+            "metadata":{
+               "name":"example-kabanero"
+            },
+            "spec":{
+               "version":"0.0.1",
+               "tekton":{
+                 "disabled":false,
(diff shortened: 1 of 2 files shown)


kyleschlosser:
I think this looks good. Let's get #1 fixed and incorporated here before we merge. 

A few other observations: 
* The schema for the spec and the status probably will change some, so we will need to revise these in the future along with the example. 
* We will need to version our Docker image rather than use 'latest'

deploy/olm-catalog/kabanero-operator/0.0.1/kabanero-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,209 @@
...
+  - 'Tekton'
+  customresourcedefinitions:
+    owned:
+    - description: Kabanero Platform Operator 
+      displayName: Kabanero Operator
+      group: kabanero.io
+      kind: Kabanero
+      name: kabaneros.kabanero.io
+      resources:
+      - kind: ConfigMap
```
kaczyns: I don't think this needs to be fixed now, but the resources should only contain the things that the operator creates.  I don't know off the top of my head what it is, but, should probably be just the three dependent CR instances that we create (pipeline, kserve, kevent).  If there are others that the operator directly creates, those can be here too.

deploy/olm-catalog/kabanero-operator/0.0.1/kabanero-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,209 @@
...
+        - description: Overrides platform component version.
+          displayName: Version
+          path: version
+          x-descriptors:
+            - 'urn:alm:descriptor:com.tectonic.ui:label'
+        - description: Specifies whether Tekton is disabled.
+          displayName: disable Tekton
+          path: tekton.disabled
+          x-descriptors:
+            - 'urn:alm:descriptor:com.tectonic.ui:booleanSwitch'
```
kaczyns: If the target environment for this is really OKD 3.x, then we will need to re-consider whether we use x-descriptors or not.  The x-descriptors work in 4.x but not 3.x.  In 4.x if you don't use a spec descriptor, it's hard to describe certain things (like true/false values) because the UI won't display true or false as a value.

PR #41, opened 2019-07-15
Update release 0.0.1 deployment yaml

mezarin:
This PR updates the version of the kabanero deployment resource image in deploy/releases/0.0.1/kabanero-operator.yaml.

### DIFF
--- a/deploy/releases/0.0.1/kabanero-operator.yaml
+++ b/deploy/releases/0.0.1/kabanero-operator.yaml
@@ -1,3 +1,736 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: knative-eventing-operator
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: knative-eventing-operator
+  template:
+    metadata:
+      labels:
+        name: knative-eventing-operator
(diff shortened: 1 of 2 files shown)


kyleschlosser:
@mezarin can you add a Kabanero namespace to that? 

```
apiVersion: v1
kind: Namespace
metadata:
  name: kabanero
```

deploy/releases/0.0.1/kabanero-operator.yaml
```diff
@@ -1,3 +1,736 @@
...
+  - name: v1alpha1
+    served: true
+    storage: true
+---
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: openshift-pipelines-operator
+  namespace: kabanero
```
kaczyns: It seems like some of these things define namespace `kabanero` and some don't define any namespace.  Should we be consistent?  I'm wondering what strange things happen if the person running this doesn't have `kabanero` as their current namespace.

jgawor: +1 ran into that as well. either specify same namespace everywhere or none at all everywhere.

PR #42, opened 2019-07-15
Add collections controller

kyleschlosser:
This pull request adds the concept of a 'collection' to the operator. When a resource of kind 'collection' is created, the controller responds by installing the related assets onto the current kubernetes cluster. 

For example, the resource: 
```
kind: Collection
metadata:
  name: java-microprofile
spec:
  version: 1.0.0
```

will pull the collection named 'java-microprofile' from the [default collections repository](https://github.com/kabanero-io/kabanero-collection/tree/master/experimental). This collection includes a number of Tekton resources. The controller will `kubectl apply` each of these resources to the current cluster. 

Additionally, a Kabanero instance can be configured globally for collections with the following additions: 
```
kind: Kabanero
metadata:
  name: kabanero
spec:
...
  collections: 
    # Installs a default set of collections
    enableFeatured: true

    # A list of those repositories which are searched for collections
    repositories: 
    - name: experimental
      url: https://github.com/kabanero-io/kabanero-collection/blob/master/experimental
```


kaczyns:
We should probably add the new CRD to the `owned` section of the CSV.  I don't know whether that happens now, or some time in the future when we have version 0.0.2 (or some other version) of the CSV.
https://github.com/kabanero-io/kabanero-operator/blob/master/deploy/olm-catalog/kabanero-operator/0.0.1/kabanero-operator.v0.0.1.clusterserviceversion.yaml

PR #50, opened 2019-07-25
Use manifestival and remove references to internal client library

kyleschlosser:
In the early draft of the operator sources, a custom client library was introduced which did client side reconciliation like `kubectl`. In some of the later additions, we started using `jcrossley3/manifestival` which has a much simpler implementation. 

This pull request consolidates the code base on the `manifestival` library and removes the original client library from the code base.


pkg/controller/kabaneroplatform/tekton.go
```diff
@@ -2,45 +2,32 @@ package kabaneroplatform
...
+	transforms := []mf.Transformer{
+		mf.InjectOwner(k),
+		mf.InjectNamespace(k.GetNamespace()),
+	}
 
-		options := &client.ApplyOptions{
-			OwningController: k,
-			Namespace:        k.GetNamespace(),
-		}
+	m.Transform(transforms...)
```
mezarin: It seems that if resource transformation did not take place, we should not be applying it. It is unclear to me whether or not an m.applyAll failure due to a transformation problem will be obfuscated; therefore, harder to debug.

stephenkinder:
License looks reasonable to use this project in Kabanero.

Issue #68, opened 2019-07-29
Collection controller always overrides repositories with default


kyleschlosser:
There is no if checking in the collection controller, so it always loads from the default: 

https://github.com/kabanero-io/kabanero-operator/blob/2881b937b2a26e7b55fa855c6a0402c9eb908c90/pkg/controller/collection/collection_controller.go#L157

Add an "if ! nil" to prevent this

stephenkinder:
@tspewak thanks for looking into this quickly.  With this bug fix allow the operator to override the default collection to point at an alternate collection?

tspewak:
I think it will.

kaczyns:
Closing as fixed by #74

PR #74, opened 2019-07-31
collection controller should not always override repositories with th…

tspewak:
…e default

### DIFF
--- a/pkg/controller/collection/collection_controller.go
+++ b/pkg/controller/collection/collection_controller.go
@@ -145,11 +145,20 @@ func (r *ReconcileCollection) ReconcileCollection(c *kabanerov1alpha1.Collection
 			}
 
 			_collection, err := SearchCollection(collectionName, index)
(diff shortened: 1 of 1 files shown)


kaczyns:
Fixes #68

pkg/controller/collection/collection_controller.go
```diff
@@ -145,11 +145,20 @@ func (r *ReconcileCollection) ReconcileCollection(c *kabanerov1alpha1.Collection
...
 			_collection, err := SearchCollection(collectionName, index)
-			r_log.Error(err, "Could not search the provided index")
+			if err != nil {
+				r_log.Error(err, "Could not search the provided index")
+			}
 			if _collection != nil {
 				collection = _collection
 			}
 		}
+                if collection != nil {
```
kaczyns: I don't understand if having multiple repositories should cause more than one collection to be created.  However this does fix the bug that was reported so lets merge this now, and we can investigate the repositories later.

Issue #67, opened 2019-07-29
Update deploy process to use new docker tag


kyleschlosser:
Update kabanero-operator/.travis/prepare_release.sh to include a sed substitution for kabanero-operator:latest -> kabaner-operator:$TRAVIS_TAG

PR #75, opened 2019-07-31
update prepare release script to update repository tag

tspewak:
The prepare release script was not putting the correct container image name in the generated yaml.  The container image should contain the tag for the release, not `latest`.


tspewak:
Fixes #67

tspewak:
the sed -i ''  version gets an error on Linux

Issue #60, opened 2019-07-29
Implement digest checking in the collection event loop


kyleschlosser:
The collections controller is a minimal implementation which when applied retrieves the remote contents and applies these contents. Any modification of the resource results in re-execution of this logic. 

Digest data is available in the remote repository for both the collection as well as the individual resources. The digest for either one or both of these should be recorded in the status, and then application of the collection be limited to the case where the digest is new. 

There is an additional case where the digests changing which may lead to re-application of the collection. This is a bit like "SNAPSHOTS" in a Maven repository, where the HEAD of the repository is changing but the version is not fully committed. This probably should not be supported as a base case, but perhaps could be recognized when a Kabanero instance configuration attribute is present. If no resolution on this is made within the current issue, we should fork this into a separate issue.

kaczyns:
The collection status was updated with the active digest, and logic to avoid re-applying the resource if the digest had not changed, in #76.  There is still the outstanding issue of whether we should attempt to apply the YAML if the digest is not correct (the computed digest on the resource does not match what is in the collection.yaml file).  I've opened #88 to track this, and I don't think it's required to close epic #66, however I don't feel strongly about this so if someone disagrees, please put #88 into epic #66.

Issue #62, opened 2019-07-29
Update the collection status


kyleschlosser:
The collection needs a status which reflects both the current state, version and digests of the collection but also any errors applying collection contents. A common error applying collection contents results from RBAC limitations within the collection contents. Some of the status recording needs to be per resource, while other parts are per collection. 

Example: 

```
status:
    # Here available version > current reflects an available update
    availableVersion: 1.0.0.1
    currentVersion: 1.0.0.0
    resources:
    - name: <resource name or path>
      currentDigest: <digest>
      status: applied
    - name: <resource name or path>
      currentDigest: <digest>
      status: failed
      statusMessage: "role" is forbidden from creating objects of kind
```

**Phases and Conditions**
Some Kubernetes objects are treated as a state machine and record their status using "phases and conditions" concepts. This allows the status to record such things as whether it is initializing, ready, failed, etc. The use of phases and conditions is advanced and optional at this point, however, some of this idea is present in the above modeling in order to account for individual resource application failures. 

For an example of this, see: 
* https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/conditions.go#L83
* https://github.com/openshift-knative/knative-eventing-operator/blob/62770a4fec0e75e81377731073c94cb3b1ca82ef/pkg/apis/eventing/v1alpha1/knativeeventing_lifecycle.go

kaczyns:
The digest and message sections of the status were resolved in PR #76.  I'd like to handle the version status in issue #61.  I'm going to close this issue.

PR #76, opened 2019-07-31
Add asset digests to the Collection status

kaczyns:
This is a partial fix for #60.  It does not address the last part about snapshots since I'm not sure I understand what the concern is yet.

The status for the Collection will now contain a list of applied assets (of type `kubernetes-resource`), and their digest.  We can then compare the digest against what is in the repository and decide whether we will re-apply the asset or not.  If the digest has changed, we'll re-apply the asset.

We use the name of the asset to find the digest in the status.  If there is no name (the example collection has no names for the kubernetes assets) then we'll use the url.

I also folded the updates from release 0.0.2 role.yaml into the CSV, since they had diverged and I got some errors about applying Tekton resources in my operator pod log.  There are still other RBAC errors that need to be dealt with when applying the resources from the example collection, because the service account for the Kabanero operator doesn't have permission to create service accounts, or to create cluster role bindings (or roles).  That needs to be a separate issue.


kaczyns:
I added a status and statusMessage as described in #62 to this PR as well.

The current behavior of the status is the controller will attempt to apply all of the kubernetes assets in the collection.  If the apply is successful, that asset is marked active, and if not it's marked failed.  If there are any failed assets, the reconcile result is set to re-queue so that it can be retried.  If there are no failed assets, no retry is scheduled.

Issue #56, opened 2019-07-26
Use docker login from stdin for Travis


kyleschlosser:
We currently do `docker login -u <u> -p <p>` but it is better to feed the password through stdin instead. See the docker login docs for an example.

stevenschader:
Investigating doc's located here -> https://docs.docker.com/engine/reference/commandline/login/

stevenschader:
example from -> https://stackoverflow.com/questions/51489359/docker-using-password-via-the-cli-is-insecure-use-password-stdin
echo "$MY_PASSWORD" | docker login --username foo --password-stdin

stevenschader:
bugger:
[0K$ if [ ! -z \"$DOCKER_USERNAME\" ]; then echo $DOCKER_PASSWORD | docker login --username $DOCKER_USERNAME --password-stdin || true; fi
Error: Cannot perform an interactive login from a non TTY device

stevenschader:
Moving the docker login out of the .travis.yml to the Makefile will work with this syntax:
docker-login:
	if [ ! -z "$(DOCKER_USERNAME)" ]; then echo $(DOCKER_PASSWORD) | docker login --username $(DOCKER_USERNAME) --password-stdin  || true; fi

push-image: docker-login
ifneq "$(IMAGE)" "kabanero-operator:latest"
	# Default push
	docker push $(IMAGE)

kaczyns:
The fix for this was not as simple as we had hoped.  I am moving it out of this iteration.

davidpelfree:
On Windows, try adding "winpty" before the "docker ..." command.
It works on GitBash, Putty and MobaXTerm.

PR #77, opened 2019-07-31
utilize the docker login --password-stdin such the password is not logged in the docker logs

stevenschader:


kaczyns:
Fixes #56

PR #85, opened 2019-08-06
add eventing-contrib

dacleyra:
Use a set-context approach in the Makefile to be able to apply namespaced and non-namespaced objects simulatenously
Update sed to a compatible approach for both MacOS and GNU

Add knative-contrib fetch to the overall get_dependencies

Pull the knative-contrib release v0.6.0. v0.7.x github releases are currently broken.

As such, change knative eventing and serving back to v0.6.0 releases for compatibility

Hardening, stop at failures
Add curl -f
set -Eeuox pipefail

### DIFF
--- a/deploy/dependencies.yaml
+++ b/deploy/dependencies.yaml
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: knative-eventing-operator
           # Replace this with the built image name
-          image: quay.io/openshift-knative/knative-eventing-operator:v0.7.0
+          image: quay.io/openshift-knative/knative-eventing-operator:v0.6.0
           command:
           - knative-eventing-operator
           imagePullPolicy: Always
@@ -222,8 +222,10 @@ spec:
       serviceAccountName: knative-serving-operator
       containers:
         - name: knative-serving-operator
-          image: quay.io/openshift-knative/knative-serving-operator:v0.7.0
-          imagePullPolicy: IfNotPresent
+          image: quay.io/openshift-knative/knative-serving-operator:v0.6.0
+          command:
+          - knative-serving-operator
+          imagePullPolicy: Always
           env:
             - name: WATCH_NAMESPACE
               value: ""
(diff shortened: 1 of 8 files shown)


deploy/dependencies.yaml
```diff
@@ -640,53 +970,72 @@ metadata:
...
+    name: status
+    type: string
   version: v1alpha1
   versions:
   - name: v1alpha1
     served: true
     storage: true
 ---
+apiVersion: operator.tekton.dev/v1alpha1
+kind: Config
```
kaczyns: My only real issue here is that there is code in the kabanero controller that tries to create an instance of the old `Install` CRD if it did not exist before.  
https://github.com/kabanero-io/kabanero-operator/blob/master/pkg/controller/kabaneroplatform/tekton.go

It seems like this code is not necessary anymore, and should be deleted, or at least verify that the `Config` object exists and log a message if it's not.

The same is also probably true of the knative-eventing and knative-serving code in the kabanero controller, as I'm not sure those objects are necessary anymore either.

I'd like to hear @kyleschlosser's comments as well as most of this code pre-dates my involvement in this project.

kyleschlosser: @dacleyra this back levels the serving and eventing operators. Can you explain why that is? 

Regarding the Tekton install CRD, we need to do two things: 
1. Update the sources so that the removed resource kind is not created or attempted
2. Remove any remaining install resource kind which we created in a prior version of the operator

dacleyra: version 0.6.0 is used instead of v0.7.x due to eventing-contrib githubsources being broken
https://github.com/knative/eventing-contrib/issues/481

The resolution to the issue has only just made it into the v0.8.0 release.

eventing-contrib 0.6 doesn't work with eventing & serving 0.7 as an option either.

Holding off to pick up newer releases, 0.7 or 0.8, dependencies and requirements are changed
https://github.com/knative/serving/releases

kyleschlosser:
We need to test a migration with this PR between the last release and this feature to see what happens. If there are old tekton objects, we may need to clean those up.

kaczyns:
I documented Kyle's issues with the code in #107, and would like to take this PR now so that we can move forward with the webhook.

Issue #80, opened 2019-08-01
Deleting a Kabanero instance does not delete Collections that it created


kaczyns:
If I delete a `Kabanero` instance, its `Collection` instance(s) are not deleted, they must be deleted separately.  We need to investigate whether it makes sense to tie the to together, similarly to how the `tekton.dev` `Install` instance is deleted when a `Kabanero` instance is deleted.

kaczyns:
The `Collection` object should be "owned by" the `Kabanero` object.  The `Kabanero` object should exist in the same namespace as the `Collection`, and there should only be one instance.  For `Collection` objects created by the Kabanero controller, it should be trivial to add this information to the `Collection` instance.  For `Collection` instances created manually (via `kubectl create -f collection.yaml`) or other, the Collection controller will need to see that there is no owner, and look up the `Kabanero` instance and set it as the owner.

This same concept already exists in other parts of the code (such as the `Install`) so it should be clear what needs to be done here.

marikaj123:
Needs review before closing it.

kaczyns:
I am closing this issue.  The main idea is finished.  The part that remains is setting the owner on manually created `Collection` objects.  This will become important when we have the CLI, but since the CLI is not ready yet, we can tackle that later.

PR #89, opened 2019-08-08
add owner reference to collection

tspewak:
Add owner reference to collection so it is deleted when kabanero is deleted.


tspewak:
Fixes #80

PR #98, opened 2019-08-09
Build break - unstructured was removed from imports.

kaczyns:
Conflict between #90 and #93.  The import of `unstructured` was removed.  I modified the code to follow the convention established by #90.


kyleschlosser:
@kaczyns how did we not catch this in the PR build? the github auto merging failed us?

kaczyns:
@kyleschlosser I don't know how Travis works... I may have created my PR before #90 merged and so I don't know how Travis would have known.  Does Travis re-submit PR builds if someone else merges something?

Issue #100, opened 2019-08-09
Set Collection owner for collections created at the CLI or manually


kaczyns:
`Collection` objects created by the Kabanero controller will have the owner reference set, so they will be deleted when the `Kabanero` object is deleted.  Those created by the CLI, will probably not have the owner set.  Those created manually by `oc create -f` or `kubectl create -f` or other means will definitely not have the owner set.

We need to investigate whether the collection controller can discover which `Kabanero` instance should be the owner, and set it during reconcile.

kaczyns:
Fixed by #102

PR #102, opened 2019-08-12
set collection owner for collections created by CLI

tspewak:
Add owner to collection objects that are created with the CLI.

### DIFF
--- a/pkg/controller/collection/collection_controller.go
+++ b/pkg/controller/collection/collection_controller.go
@@ -10,7 +10,9 @@ import (
 	kabanerov1alpha1 "github.com/kabanero-io/kabanero-operator/pkg/apis/kabanero/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


tspewak:
Fixes #100

pkg/controller/collection/collection_controller.go
```diff
@@ -191,6 +235,13 @@ func (r *ReconcileCollection) ReconcileCollection(c *kabanerov1alpha1.Collection
 	}
 	r_log = r_log.WithValues("Collection.Name", collectionName)
 
+	// A collection created by the CLI might not have kabanero as the owner.
+	// In that case we want to make kabanero the owner
+	err := r.ensureCollectionHasOwner(c, k)
+	if err != nil {
+		r_log.Error(err, "Could not make kabanero the owner of the collection")
```
kaczyns: Could we update the error message with the name of the collection (collectionName)?  I know we trace it just before, but I don't know if it's possible to configure the log to just print error messages, in which case we wouldn't have the collectionName.

tspewak: The r_log.WithValues("Collection.Name", collectionName) just above cause the collection name to come out on the r_log.Error(err, "Could not make kabanero the owner of the collection")

kaczyns: Thanks Tim.  I didn't understand how the log worked.  I will merge this.

Issue #31, opened 2019-07-10
Kabanero CSV owned CRD resources is not correct


kaczyns:
The kabanero-operator CSV says that the `Kabanero` CRD owns the following resources:
```yaml
      resources:
      - kind: ConfigMap
        name: ""
        version: v1
      - kind: Deployment
        name: ""
        version: v1
      - kind: Pod
        name: ""
        version: v1
      - kind: Role
        name: ""
        version: v1
      - kind: RoleBinding
        name: ""
        version: v1
      - kind: ServiceAccount
        name: ""
        version: v1
```
This is not correct - the Kabanero operator creates a Tekton `Install`, a Knative `KnativeEventing` and a Knative `KnativeServing`.  

The OKD / OpenShift console uses the `resources` section as a clue for what to display when viewing an instance of our CRD.  It's not broken the way it is, but it's not useful either.

kaczyns:
The `Kabanero` instance also creates a `Collection` instance, and we should consider linking the two together.

kaczyns:
Time has passed and the "owned" resources continue to evolve.  This is where I think we stand:
1)  A `Kabanero` instance creates a `Collection` - it should own that.  The collection object should have its owner set now so it should be displayed correctly.
2) A `Collection` doesn't directly own anything, however it does create other resources based on what it reads from the collection yaml.  So far we have seen Tekton `Pipeline` and `Task` objects.  These should also have their owner set and we can include those in the list of things the `Collection` owns.
3) As we see other collections join the repository, we'll have to look at what kinds of resources they create, and add them to the owned list as well.

PR #105, opened 2019-08-14
add resources to owned collection In CSV

tspewak:
Add pipeline ant task resources to the owned collection in the kabanero CSV


kaczyns:
fixes #31

Issue #63, opened 2019-07-29
Allow featured collections installation to be specified on a per repository basis


kyleschlosser:
Currently whether to install the featured collections is on a per Kabanero instance basis, but this should ideally be modeled as per repository. Featured collections are to be installed by default -- if the configuration value is unspecified, then featured collection installation is enabled. 

The existing repository modeling is: 
```
repositories: 
- name: myrepo
  url: http://....
```

Proposed: 
```
repositories: 
- name: myrepo
  url: http://....
  installFeaturedCollections: true
```

kaczyns:
After talking with @stephenkinder today, it sounds like "featuredCollections" refers to the default collections provided by Kabanero.  Any additional collections would be provided by URLs in the `repositories` section.  There is no instance of a "featured" collection in these repositories.

The current behavior should be that if "featuredCollections" are enabled, then the default Kabanero collections are installed.  Then, any repositories are processed and all collections in those repositories are installed.

@kyleschlosser please let us know you have a different thought.  It sounds like `installFeaturedCollections` on a specific repository is not necessary.

kaczyns:
@tspewak and I spoke with @kyleschlosser today.  My understanding of what a featured collection is, is incorrect.  It should be possible to mark a collection within an index (repository) as 'featured'.  When a `Kabanero` instance is created, the spec should indicate whether or not featured collections are automatically activated (featuredCollections=true/false).  If they are automatically activated, then a `Collection` object is created for each collection in the repository that is marked 'featured'.  If they are not automatically activated, then no `Collection` objects are created, and the individual collections must be activated manually, either by creating a `Collection` object using `kubectl` or `oc`, or by using proposed CLI tools which are still under development.

The current state of the kabanero_controller is correct - in that if "featuredCollections" is true, then all of the collections in the default index are created.  This is because the controller makes an assumption that all of the collections are featured.  This assumption was not communicated well.

This issue should move the `enableFeatured` attribute of the `Kabanero` Spec.Collections section, into the Spec.Collections.Repositories section, so that `enableFeatured` can be enabled per-repository.  We should validate that the `Kabanero` controller correctly processes multiple repositories where `enableFeatured` is enabled for one, but not for the other.

[...]

PR #106, opened 2019-08-15
Allow featured collection install to be per repository

tspewak:
Remove enableFeatured and add activateDefaultCollections at the repository level so featured collections can be installed per repository.


tspewak:
Fixes #63

Issue #91, opened 2019-08-08
Remove use of the default collection repository from collection controller


kaczyns:
The collection controller will currently use the default URL to find the default collection repository, if no URL is provided on the `Kabanero` object.  This is not appropriate, since in a production environment it might be an unwelcome surprise to access an unknown URL if a mistake is made while coding the `Kabanero` instance.

PR #122, opened 2019-08-23
Use version 0.1.0 of the Kabanero collections

kaczyns:
The collections team released 0.1.0 and we're updating our samples to use that.


kaczyns:
Thanks for pointing that out @dacleyra - I think that #91 will be merged soon and that should remove that reference.  But in case that does not happen.... we'll need to fix it.

PR #123, opened 2019-08-23
91--Collective controller to not use a default repo if none supplied

ginnick:
Update the collective controller to not default to a index.yaml if no repository is defined in the Kabanero instance.

### DIFF
--- a/pkg/controller/collection/collection_controller.go
+++ b/pkg/controller/collection/collection_controller.go
@@ -258,13 +258,15 @@ func (r *ReconcileCollection) ReconcileCollection(c *kabanerov1alpha1.Collection
 	}
 
 	// Retreive all matching collection names from all remote indexes.  If none were specified,
-	// use the default.
+	// update the status and return.
 	var matchingCollections []resolvedCollection
 	repositories := k.Spec.Collections.Repositories
-	if len(repositories) == 0 {
-		default_url := "https://raw.githubusercontent.com/kabanero-io/kabanero-collection/master/experimental/index.yaml"
-		repositories = append(repositories, kabanerov1alpha1.RepositoryConfig{Name: "default", Url: default_url})
-	}
+        if len(repositories) == 0 {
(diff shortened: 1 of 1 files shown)


ginnick:
Fixes #91

ginnick:
Here's what it looks like in the Pod's log:

{"level":"info","ts":1566582904.4861455,"logger":"controller_collection","msg":"No repositories specified in collection (java-microprofile)","Request.Namespace":"kabanero","Request.Name":"java-microprofile","Collection.Name":"java-microprofile"}

pkg/controller/collection/collection_controller.go
```diff
@@ -258,13 +258,15 @@ func (r *ReconcileCollection) ReconcileCollection(c *kabanerov1alpha1.Collection
...
-	// use the default.
+	// update the status and return.
 	var matchingCollections []resolvedCollection
 	repositories := k.Spec.Collections.Repositories
-	if len(repositories) == 0 {
-		default_url := "https://raw.githubusercontent.com/kabanero-io/kabanero-collection/master/experimental/index.yaml"
-		repositories = append(repositories, kabanerov1alpha1.RepositoryConfig{Name: "default", Url: default_url})
-	}
+        if len(repositories) == 0 {
+		r_log.Info(fmt.Sprintf("No repositories specified in collection (%s)",collectionName))
```
kaczyns: I think we want to report that there are no repositories configured in the Kabanero instance, not in the collection.  The Kabanero instance is where we're looking for repositories (`k.Spec.Collections.Repositories`).  And I agree with what @tspewak and @mezarin said about this being an error message instead of Info, even though it doesn't look like they wrote that down anywhere :-)

kaczyns: you should be able to do a `r_log.Error()` but you'll need an Error object to pass.  
```go
errors.New(fmt.Sprintf("Your message here"))
```
Then it's something like:
```go
r_log.Error(err, "More info about the error")
```

Issue #113, opened 2019-08-21
Expose namespace to CLI microservice


davco01a:
Expose namespace(s) to CLI microservice so it knows which namespaces to activate collections into when performing refresh

davco01a:
configMap?

kaczyns:
I am pretty sure we can include an `env:` statement in the template for the pod that has an environment variable defined which you could look up with a `System.getenv()`.  I don't think we'll need a config map.  I need to show you how to do it but I am not able to switch to the branch that has that change right now.

PR #124, opened 2019-08-23
Cli deployment

### LINKED ISSUE #86: Start the "Kabanero CLI" backend microservice
The Kabanero CLI back-end is being developed here:
https://github.com/kabanero-io/kabanero-command-line-services

The Kabanero operator needs to deploy this microservice as a Kubernetes deployment, when the container hosting the microservice is available on kabanero Dockerhub.

kaczyns:
Fixes #86 and #113 
The Kabanero CLI microservice is deployed.  It's a container containing Liberty.  It is probably not a container based on the UBI image but that can be addressed as a follow-on when we go thru certification processing.  Some of the configuration is taken from a config map, which we've populated with sample data.  This probably needs to be more flexible going forward.  A route is created and its hostname is copied to the Kabanero Status so that it's in a convenient place for tooling to find.


kaczyns:
@kyleschlosser Yes there is no logic for Route - I thought that OKD and OpenShift were our target environments.  If we are trying to run on other environments, there are other changes that we'll need to make.  I think the install script installs the open shift version of Tekton.  Also the Istio instructions assume OKD or OpenShift.

PR #143, opened 2019-08-29
Allow the Github organization to be specified on the Kabanero object

### LINKED ISSUE #138: need environment variable for group/org in deployment yaml
need environment variable for group/org in deployment yaml

Currently we have this hardcoded in the microservice but it needs to be portable for customer's environments

kaczyns:
Fixes #138.
The Github organization is now specified on the Kabanero Spec:
```yaml
spec:
  version: 0.0.1
  githubOrganization: production
  collections: 
```

When `githubOrganization` is specified, the container specification for the CLI will contain a new environment variable, `KABANERO_CLI_GROUP` which will be set to whatever value was specified (in this case `production`).

I tested that the environment variable is present when specified, and not present when not specified.  I also tested that when the value of `githubOrganization` changes (ie applying changes to an existing `Kabanero`) that the deployment also changes, and the pod recycles with the new value (or no value if it was removed).

Feel free to suggest a new name or structure for the `Spec` section - I am open to using something else.


pkg/controller/kabaneroplatform/deployments.go
```diff
@@ -0,0 +1,139 @@
+package kabaneroplatform
```
mezarin: This looks ok, but here are some suggestions to handle now or in a future PR:
1. Since i foresee us adding many more resource types, it would be better if we put this file into it's own dir/package (i.e. kabaneroplatform/resources).
2. getOwnerReference() should be either public (or not in same package) or it should be in it's own Utils type part because i also see being used by other types.
3. In my opinion, we should have a template deployment and setters that callers can use. The setters would handle default/invalid values.

kyleschlosser:
This needs to be reevaluated after discussion with the originating requestor

Issue #151, opened 2019-09-04
Loop when Kabanero instance refers to a non existing index


kyleschlosser:
Related to #147, if the Kabanero instance refers to a URL which produces a 404, we need to log this in the status and do not requeue the request, or extend the requeue time. Requeuing immediately causes a forever loop.

PR #180, opened 2019-09-10
avoid tight loop when repository url is bad

tspewak:
Instead of requesting a requeue every 60 seconds when the repository url is bad, the first requeue will be in 60 seconds. After the first requeue the requeue time increases by 60 seconds up to a max of 15 minutes.  Kabaneros status is updated with the error that occurred.

### DIFF
--- a/pkg/controller/kabaneroplatform/kabaneroplatform_controller.go
+++ b/pkg/controller/kabaneroplatform/kabaneroplatform_controller.go
@@ -3,6 +3,7 @@ package kabaneroplatform
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


tspewak:
Fixes #151

pkg/controller/kabaneroplatform/kabaneroplatform_controller.go
```diff
@@ -33,7 +34,7 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileKabanero{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileKabanero{client: mgr.GetClient(), scheme: mgr.GetScheme(), requeueDelayMapV1: make(map[string]RequeueData), requeueDelayMapV2: make(map[string]RequeueData)}
```
kaczyns: I am not thrilled by how much code we're having to duplicate for the V1 and V2 collections.  Since the V1 collection was such a small blip in time, I'll open an issue to resolve this (either remove support for V1, or come up with another abstraction that doesn't require us to duplicate code).

PR #186, opened 2019-09-16
182 - Watch Collections and Collection objects

dacleyra:
- Add the scheme for tekton
- Create a predicate for Collection, so that reconcile does not occur on an update event with no generation change
- Add Watches for tekton objects from collection, add a predicate such that it doesn't reconcile with no generation changes, or when the object is created - otherwise requeues keep occuring since the create occurs from the Collection watch
- Don't delete a collection then Apply, just Apply
- Don't prevent the collection from being applied just because the archive status.sha matches the index sha, the docs may have been deleted or modified (older logic from initial v1 kab index)
- KabaneroPlatform Controller Watch Collections


cmd/manager/main.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
+	
+	pipelinev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
```
kaczyns: Does this package need to be copied to the vendor directory?

pkg/controller/collection/collection_controller.go
```diff
@@ -44,19 +47,54 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	}
 
+	// Create Collection predicate
+	c_pred := predicate.Funcs{
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// Ignore updates to CR status in which case metadata.Generation does not change
+			return e.MetaOld.GetGeneration() != e.MetaNew.GetGeneration()
```
kaczyns: This is really powerful, and I didn't realize what it was doing until I searched the doc for `.metadata.generation`.  Basically this is what prevents the reconciler from running again when the operator updates the status of something.  Would you mind changing the comment so it says this more explicitly?  Something like "Returning true only when the metadata generation has changed, allows us to ignore events where only the object status has changed, since the generation is not incremented when only the status changes".  Feel free to ignore this comment unless you're in here to change other things.

pkg/controller/collection/collection_controller.go
```diff
@@ -652,6 +690,8 @@ func activatev2(collectionResource *kabanerov1alpha1.Collection, collection *Ind
 			transformedManifests = append(transformedManifests, transformedRemoteManifest{m, asset.Url})
 		}
 
+// 182: Do not delete, only apply
+/*
```
kaczyns: I am not a big fan of commenting out code that's under version control.  If we've decided that we're not going to do it, I would prefer to just delete it, because someone can come back later and look at the prior version in Git to see what it used to look like.  Again, feel free to ignore this comment if others disagree.

cmd/manager/main.go
kaczyns: It looks like the Travis build failed, I think that this does need to be vendored.

dacleyra: added new vendor files
https://github.com/kabanero-io/kabanero-operator/pull/186/commits/250ac65664e5791515c914d4ca5ab812591a0702

pkg/controller/collection/collection_controller.go
dacleyra: Ok, will clean that up

pkg/controller/collection/collection_controller.go
dacleyra: ok, will do

dacleyra: comment updated in commit https://github.com/kabanero-io/kabanero-operator/pull/186/commits/d71fb952c78211209a2063be652e0ab60d62e2ac

pkg/controller/collection/collection_controller.go
dacleyra: removed commented code in  commit https://github.com/kabanero-io/kabanero-operator/pull/186/commits/d71fb952c78211209a2063be652e0ab60d62e2ac

PR #190, opened 2019-09-18
Add watch additional namespaces

kyleschlosser:
Under the current design, a Kabanero instance is associated with the namespace in which the Kabanero resource is created. Note that although our examples may place the Kabanero resource into the namespace `kabanero` this is unnecessary, the Kabanero resource can go into any valid namespace within the cluster. This design means that Appsody and KAppNav are both associated only to the namespace that hosts the Kabanero instance and collections, however, it's common for applications developed through Kabanero to be deployed into other namespaces. 

This pull request adds the ability for Kabanero to be hosted in one namespace (the meta/namespace associated with the Kabanero instance resource) but additionally watch for applications in other namespaces. The additional namespaces are expressed in the Kabanero resource like this: 

```
kind: Kabanero
metadata:
  name: kabanero
spec:

  targetNamespaces:
  - ns1
  - ns2
```


stephenkinder:
This PR is an excellent start @kyleschlosser.  A couple of questions: 

1.  Who is responsible for creating the namespace listed in the Kabanero CR, and if not Kabanero, I presume the namespace must exist before any Kabanero operation.  I may have missed the code that would create the namespace.  

1.  How does the developer know what namespaces are valid?  If I understand the issue, the appsody_deploy.yaml named an arbitrary target namespace; which suggests that somehow the developer knew the valid set of physical namespaces?  I guess my preference would be to have Kabanero be able to tolerate them being there, or create them if they don't exist.

1.  What is the default if none are specified?  We could treat the existing behavior as a `bug` and `correct` the behavior by watching for all namespaces and assume that this Kabanero instance has the whole cluster to itself.

1.  What happens when the developer specifies an unmanaged namespace in the appsody_deploy.yaml; i.e. the namespace is not specified in the Kabanero Instance CR.   Are we expecting that other features of Kabanero would validity check the set of target namespaces?  

1.  By default, I suppose pipelines always run in the Kabanero Instance namespace.  Just wondering if the pipelines really should be running in the `kabanero` instance namespace itself or if pipelines should default to their own namespace.

1.  Would we ever expect that a single Kabanero instance would manage two discrete sets of target namespaces?  For example, if two teams use the same Kabanero instance, would we expect microservice build, tests and deployments to be loosely managed across the same set of namespaces, or would we expect that Kabanero would help direct actions to independent sets of namespaces?    My suggestion is to leave the opportunity for Kabanero to manage independent sets of namespaces, but not implement that capability.  A simple way might be to adjust the Kabanero CRD yaml schema to allow for multiple lists of namespaces; but only implement for one set.

davco01a:
@kyleschlosser @stephenkinder @kaczyns Is this slated for the October 7th release?

stephenkinder:
@davco01a yes; not having Kabanero managed namespaces is a big problem for real application deployments.

kyleschlosser:
> This PR is an excellent start @kyleschlosser. A couple of questions:
> 
> 1. Who is responsible for creating the namespace listed in the Kabanero CR, and if not 
Kabanero, I presume the namespace must exist before any Kabanero operation.  I may have missed the code that would create the namespace.  I guess my preference would be to have Kabanero be able to tolerate them being there, or create them if they don't exist.

A security challenge with this is that it could be viewed as a "privilege escalation" flaw where RBAC authority to `edit` any Kabanero in the cluster would supercede the lack of `create` for namespaces. That is, if you could edit any given Kabanero, you could create namespaces across the cluster at will by just listing them out in any of the Kabanero instances. We would need to address that concern before adding such a feature. 

> 2. How does the developer know what namespaces are valid?  If I understand the issue, the appsody_deploy.yaml named an arbitrary target namespace; which suggests that somehow the developer knew the valid set of physical namespaces?

The namespace being deployed to was well known in the case which led to this enhancement request. The workspace concept is probably better suited to the case where these are not well known. 

> 3. What is the default if none are specified?  We could treat the existing behavior as a `bug` and `correct` the behavior by watching for all namespaces and assume that this Kabanero instance has the whole cluster to itself.

This is an extension to the current behaviors rather than a change. If the newly modeled attribute is unset, there is no difference from the prior behavior, making the proposed change backward compatible. 

> 4. What happens when the developer specifies an unmanaged namespace in the appsody_deploy.yaml; i.e. the namespace is not specified in the Kabanero Instance CR.   Are we expecting that other features of Kabanero would validity check the set of target namespaces?

Anything that happens outside the jurisdiction of the Kabanero instance is unmanaged by the current Kabanero instance, and this is implicit in the design of namespace scoping. That is, we do not require all namespaces in the cluster to be managed by a Kabanero instance, and therefore you cannot assert for a given resource if a user meant for this to belong to a Kabanero instance or if they mean for it to be unmanaged. 

> 5. By default, I suppose pipelines always run in the Kabanero Instance namespace.  Just wondering if the pipelines really should be running in the `kabanero` instance namespace itself or if pipelines should default to their own namespace.

By convention, created resources belong to the same namespace as the resource which caused them to be created. By this rationale, the kabanero instance and the collections belong to the same namespace. If you want pipelines in another namespace, that would need to be explicitly modeled. 

> 6. Would we ever expect that a single Kabanero instance would manage two discrete sets of target namespaces?  For example, if two teams use the same Kabanero instance, would we expect microservice build, tests and deployments to be loosely managed across the same set of namespaces, or would we expect that Kabanero would help direct actions to independent sets of namespaces?    My suggestion is to leave the opportunity for Kabanero to manage independent sets of namespaces, but not implement that capability.  A simple way might be to adjust the Kabanero CRD yaml schema to allow for multiple lists of namespaces; but only implement for one set.

There is no 'team' concept in this proposal, but if you were to do team scoped namespace sets, you would model it quite differently from what I have proposed here. I would suggest that if you want such a capability, we put that idea into a new enhancement.

stephenkinder:
> > This PR is an excellent start @kyleschlosser. A couple of questions:
> > 
> > 1. Who is responsible for creating the namespace listed in the Kabanero CR, and if not
> >    Kabanero, I presume the namespace must exist before any Kabanero operation.  I may have missed the code that would create the namespace.  I guess my preference would be to have Kabanero be able to tolerate them being there, or create them if they don't exist.
> 
> A security challenge with this is that it could be viewed as a "privilege escalation" flaw where RBAC authority to `edit` any Kabanero in the cluster would supercede the lack of `create` for namespaces. That is, if you could edit any given Kabanero, you could create namespaces across the cluster at will by just listing them out in any of the Kabanero instances. We would need to address that concern before adding such a feature.
> 
Security issue is good point.  I have no issue with requiring the namespaces be created _a priori_.  That said, the operator should report that a specified namespace does not exist, possibly in the status field? 
> > 1. How does the developer know what namespaces are valid?  If I understand the issue, the appsody_deploy.yaml named an arbitrary target namespace; which suggests that somehow the developer knew the valid set of physical namespaces?
> 
> The namespace being deployed to was well known in the case which led to this enhancement request. The workspace concept is probably better suited to the case where these are not well known.
> 
Understood.  The enterprise architect, kubernetes administrator, and development must have agreed to this namespace _a priori_.   I could agree that Kabanero can do better with a new concept in the future.
> > 1. What is the default if none are specified?  We could treat the existing behavior as a `bug` and `correct` the behavior by watching for all namespaces and assume that this Kabanero instance has the whole cluster to itself.
> 
> This is an extension to the current behaviors rather than a change. If the newly modeled attribute is unset, there is no difference from the prior behavior, making the proposed change backward compatible.
> 
I wasn't concerned about backwards compatibility; just wondering if we should just require the appsody operator to be cluster-wide if the namespace watch list is not specified in the Kabanero CR.  Documentation?
> > 1. What happens when the developer specifies an unmanaged namespace in the appsody_deploy.yaml; i.e. the namespace is not specified in the Kabanero Instance CR.   Are we expecting that other features of Kabanero would validity check the set of target namespaces?
> 
> Anything that happens outside the jurisdiction of the Kabanero instance is unmanaged by the current Kabanero instance, and this is implicit in the design of namespace scoping. That is, we do not require all namespaces in the cluster to be managed by a Kabanero instance, and therefore you cannot assert for a given resource if a user meant for this to belong to a Kabanero instance or if they mean for it to be unmanaged.
> 
I would have expected that we'd want some validation at some point; perhaps that could just be a warning, and the real failure comes if you haven't: created the namespace and setup the apposdy operator to listen for appsody deployments in that namespace.  
> > 1. By default, I suppose pipelines always run in the Kabanero Instance namespace.  Just wondering if the pipelines really should be running in the `kabanero` instance namespace itself or if pipelines should default to their own namespace.
> 
> By convention, created resources belong to the same namespace as the resource which caused them to be created. By this rationale, the kabanero instance and the collections belong to the same namespace. If you want pipelines in another namespace, that would need to be explicitly modeled.
> 
I think this is a fine default, but I think we should document this behavior and make it more clear -- Kabanero collection pipelines run in the Kabanero instance namespace.
> > 1. Would we ever expect that a single Kabanero instance would manage two discrete sets of target namespaces?  For example, if two teams use the same Kabanero instance, would we expect microservice build, tests and deployments to be loosely managed across the same set of namespaces, or would we expect that Kabanero would help direct actions to independent sets of namespaces?    My suggestion is to leave the opportunity for Kabanero to manage independent sets of namespaces, but not implement that capability.  A simple way might be to adjust the Kabanero CRD yaml schema to allow for multiple lists of namespaces; but only implement for one set.
> 
> There is no 'team' concept in this proposal, but if you were to do team scoped namespace sets, you would model it quite differently from what I have proposed here. I would suggest that if you want such a capability, we put that idea into a new enhancement.
I wasn't asking us to solve the team concept in this PR; i was just asking that we architect the schema of this change to be more flexible.  The current schema limits to one list.  

All this said, I think the operator squad is likely better to review the code.  I have taken a pass through it, but they likely will give you a more quality review than I.

davco01a:
Connecting to our Epic, please connect to other relevant Epics too

kaczyns:
Testing went OK.  I had to use `daily` appsody-operator.  When there is a new named release I'll update the version struct to use it, and then merge this in.  Outlook is Sept 30th for a new release of the appsody operator.

arthurdm:
hey @kaczyns - would it be possible for Kabanero to pull in a particular version of the Appsody Operator, but tolerate patch updates?   For example, our next planned release is `0.2.0`, but let's say a week after Kabanero GA's we find a bug in the operator and release `0.2.1`.  It would be nice if Kabanero automatically picked up that new patch release in subsequent installs.   

Essentially the ask is:  can kabanero install `0.2.x`?

kaczyns:
@arthurdm @navidsh @leochr unfortunately not right now.  The kabanero operator has copies of the YAMLs and so wouldn't recognize a new version.  I have a couple of ideas....
1) The user can over-ride the appsody operator image in the `kind: Kabanero` instance.  If something critical came out we could give instructions for how someone could over-ride the image.  I would probably not prefer this as it's on the user to do the over-riding.
2) If you decided to have a docker image like `0.2.x` or `0.2` which was meant to be the latest release of 0.2.x of appsody operator, and if you were comfortable with us pointing at that, we could point to that instead.  That way, if you had a new patch release, in addition to creating `0.2.1` for example, you could also update `0.2` or `0.2.x` and then the next time the appsody operator pod was created, it would pull down the latest patch release.  I guess personally I would want to know exactly the level the customer was running at (ie Kabanero 0.2.0 always uses appsody operator 0.2.0) but you'll have to tell me what you prefer.

arthurdm:
moving the appsody operator discussion out of this PR and into https://github.com/kabanero-io/kabanero-operator/issues/204 (sorry for the sidetrack comments!)

PR #195, opened 2019-09-19
add rbac for landing page

alohr51:
There is a new landing page coming that will show information using the Kubernetes API such as Kabanero instance name, date created, collection hub URL, and Collections installed. To do this the deployment will need some permissions to access resources via the API.

part of https://github.com/kabanero-io/kabanero-foundation/issues/75

Give the landing page permissions to:
* read/list/watch routes 
* read/list/watch routes/status
* read/list/watch all resources in the kabanero.io apigroup

These are cluster roles (for now) due to the fact there could be multiple kabanero instances in the future (in different namespaces).

This could open a discussion to whether we want a kabanero landing application for each instance or 1 for all instances.

Tested with a custom operator on OKD 3.1.1

### DIFF
--- a/config/orchestrations/landing/0.1/kabanero-landing.yaml
+++ b/config/orchestrations/landing/0.1/kabanero-landing.yaml
@@ -25,3 +25,52 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: kabanero-landing
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kabanero-landing
+subjects:
+- kind: ServiceAccount
+  name: kabanero-landing
+  namespace: kabanero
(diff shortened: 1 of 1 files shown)


alohr51:
There is a new image for the landing page that will be ready around tues next week, it will be available via the kabanero dockerhub like the current one is today, but with a new tag.

kaczyns:
@kyleschlosser I wanted to get your opinion on this since you introduced the versioning code.  Not on the new role - I think that's fine.  How would you classify this - would you add it to the v1 orchestration for the landing page, or add a v2 at this point?  It's more complicated than just changing the image / tag.  If I picked, I'd pick v2 since there are new resources, but I wanted to hear your thoughts.

kaczyns:
... and now that I re-read the PR I see you've already brought that up :-) just ignore me....

PR #210, opened 2019-09-30
Remove KAppNav from the operator in preparation for inclusion in the install script

kaczyns:
It was decided that KAppNav should be installed at the cluster level, therefore it's not appropriate for it to be included in the Kabanero operator which currently installs things at the namespace level.  In preparation for moving it to the install script, we've removed its reconciler from the operator.  

I've also fixed a bug where the `manifest.yaml` in the collection didn't contain any checksums (as was the case in version 0.1.2 of the collections) and the collection controller would treat this as an error.  The controller now just skips the checksum validation for that file (or files).


pkg/controller/kabaneroplatform/kappnav.go
```diff
@@ -19,137 +15,36 @@ var kanlog = rlog.Log.WithName("kabanero-kappnav")
...
+		Kind:    "KAppNav",
+		Group:   "charts.helm.k8s.io",
+		Version: "v1alpha1",
+	})
+	err := c.Get(context.Background(), client.ObjectKey{
+		Namespace: "kappnav",
+		Name:      "instance",
+	}, u)
+	
+	if err == nil {
```
mezarin: Just one comment. Checking that the instance exists to report that KAppNav is ready may not reflect the actual readiness of KAppNav.

kaczyns: I will open up a follow-on issue to give a better status report.  I will look at your PR that we canceled, to see how you were reading the status etc from the KAppNav CR instance.

Issue #188, opened 2019-09-17
Expose JwtExpiration as environment variable via configMap


davco01a:
Expose JwtExpiration as environment variable via configMap

Add JwtExpiration as a number of minutes to configmap and env var via Kabanero Document

kaczyns:
@davco01a @brutif We should have a conversation about where this belongs in the Kabanero document.  The things we've added so far were directly related to our interaction with Github, and so they were in a `Github:` section of the Kabanero document.  How do we explain to the user what this is?  I am guessing they are not going to know what `JwtExpiration` is or how it affects what they are doing.  I'm assuming that this is the amount of time that a user's session with the CLI will remain until they have to log in again.  Do you have a recommendation as to how we can add this to the Kabanero document in a way that the user will understand?

davco01a:
KabaneroCLILoginExpiration may be a better name

davco01a:
Agree that this is broader than just github, but documenting shouldn't be too hard, we just have to find a suitable spot in the Kabanero document

brutif:
@kaczyns @davco01a can you point me to the document please?

kaczyns:
@brutif the document @davco01a is referring to is the Kabanero CRD:
https://raw.githubusercontent.com/kabanero-io/kabanero-operator/master/deploy/crds/kabanero_kabanero_crd.yaml

There are several `spec:` sections - the section that has `properties:` and then `github:` is the one we are referring to.  There is a `cliServices:` section that was recently added, maybe that's where this goes.  Something like `spec.cliServices.sessionExpirationMinutes` maybe....

brutif:
@kaczyns thanks.  I was looking at https://github.com/kabanero-io/docs/blob/master/ref/general/installing-kabanero-foundation.adoc and not seeing anything about editing any kind of config file, which might still be a gap. Whatever you and @davco01a think is a good name is fine, adding detection of the environment variable to the app is trivial once we know it's name.

davco01a:
`cliServices.sessionExpirationMinutes` sounds good to me

davco01a:
we'll need to get this into into:  https://github.com/kabanero-io/docs/issues/66

davco01a:
directly into env var is fine

dacleyra:
closing, resolved by https://github.com/kabanero-io/kabanero-operator/pull/213

PR #213, opened 2019-10-01
Expose JwtExpiration

dacleyra:
Results in envVar set

Question of whether the yaml field should be a string or int?

```
oc describe pod kabanero-cli-7f7688cbd8-jjqj6
Name:               kabanero-cli-7f7688cbd8-jjqj6
Namespace:          kabanero
Priority:           0
PriorityClassName:  <none>
Node:               annal3.fyre.ibm.com/172.16.20.219
Start Time:         Tue, 01 Oct 2019 11:00:54 -0400
Labels:             app=kabanero-cli
                    pod-template-hash=3932447684
Annotations:        openshift.io/scc=restricted
Status:             Running
IP:                 10.130.0.254
Controlled By:      ReplicaSet/kabanero-cli-7f7688cbd8
Containers:
  kabanero-cli:
    Container ID:   docker://84eaf21284bc5e16bf61c09735c48b7d3b092d23b4f3d73ea3586488a698f650
    Image:          kabanero/kabanero-command-line-services:0.1.1-rc.1
    Image ID:       docker-pullable://docker.io/kabanero/kabanero-command-line-services@sha256:9f569d24d1651f4a29aea44b67232a0d288e8a8ba4e7dda3ffc66837660e6601
    Port:           9443/TCP
    Host Port:      0/TCP
    State:          Running
      Started:      Tue, 01 Oct 2019 11:01:09 -0400
    Ready:          True
    Restart Count:  0
    Environment:
      KABANERO_CLI_NAMESPACE:  kabanero
      KABANERO_CLI_GROUP:      kabanero-io
      teamsInGroup_admin:      adminTeam@kabanero-io
      github.api.url:          https://api.github.com
      JwtExpiration:           60
      AESEncryptionKey:        <set to the key 'AESEncryptionKey' in secret 'kabanero-cli-aes-encryption-key-secret'>  Optional: false
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kabanero-cli-token-lm5rj (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             True
  ContainersReady   True
  PodScheduled      True
Volumes:
  kabanero-cli-token-lm5rj:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  kabanero-cli-token-lm5rj
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  node-role.kubernetes.io/compute=true
Tolerations:     <none>
```

### DIFF
--- a/config/samples/full.yaml
+++ b/config/samples/full.yaml
@@ -40,6 +40,12 @@ spec:
     # Overrides the image uri
     image: kabanero/kabanero-command-line-services:0.1.1-rc.1
 
+    # Indicates the token expiration time 
+    # Specify a positive integer followed by a unit of time, which can be hours (h), minutes (m), or seconds (s). 
+    # For example, specify 30 seconds as 30s. 
+    # You can include multiple values in a single entry. For example, 1m30s is equivalent to 90 seconds.
+    sessionExpirationSeconds: "1440m"
+
   landing:
     # The landing page is enabled by default. To disable specify false.
     enable: true 
--- a/pkg/apis/kabanero/v1alpha1/kabanero_types.go
+++ b/pkg/apis/kabanero/v1alpha1/kabanero_types.go
@@ -81,6 +81,7 @@ type KabaneroCliServicesCustomizationSpec struct {
 	Image      string `json:"image,omitempty"`
 	Repository string `json:"repository,omitempty"`
 	Tag        string `json:"tag,omitempty"`
+	SessionExpirationSeconds string `json:"sessionExpirationSeconds,omitempty"`
 }
 
 type KabaneroLandingCustomizationSpec struct {
--- a/pkg/controller/kabaneroplatform/cli.go
+++ b/pkg/controller/kabaneroplatform/cli.go
@@ -19,6 +19,7 @@ import (
 	"math/big"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
+	"regexp"
 )
 
 func reconcileKabaneroCli(ctx context.Context, k *kabanerov1alpha1.Kabanero, cl client.Client, reqLogger logr.Logger) error {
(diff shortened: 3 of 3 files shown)


davco01a:
Hmm, we need to get this populated into the server config in the same manner as we did for AESEncryptionKey:

```
<jwtBuilder expiresInSeconds="86400" id="kabsecbuilder" issuer="${jwt.issuer}" 
        keyStoreRef="defaultKeyStore" keyAlias="default" signatureAlgorithm="HS256" 
        sharedKey="${AESEncryptionKey}"/>
```

davco01a:
Ok, I have to add the env var to the server xml of the microservice

kaczyns:
> Looks fine, we can always convert to int if necessary, thanks Dan

@davco01a are you saying that you're comfortable handling the error in the CLI if someone types `FRED` into the field?  What is the CLI going to do in this case?

dacleyra:
@kaczyns bad input should be handled now with regex check

kaczyns:
it seems like `time.ParseDuratin(string)` may help with the validation part of things if we decide to go this way (supports parsing things like 2h30m50s etc) and will return an error message if it doesn't parse.  We'd still have to check that it was a positive number.

kaczyns:
Fixes: #188

davco01a:
@kaczyns Hmm FRED minutes, probably not going to work

davco01a:
maybe we should just keep it simple, and drop back to just specifying a number of minutes.  Let it be a string entered, and then in the code concatenate an "m" to it

davco01a:
I'll let @kaczyns merge since he has a change request

kaczyns:
@dacleyra and I spoke about this offline... I did not realize that the `sessionExpireMinutes` is being fed directly to the Liberty JWT configuration variable `expireInSeconds`.... which is a bit odd.  Even odder since `expireInSeconds` can actually take other duration types (so for example the value could be 3m30s).  It's not intuitive but that's really beside the point because the Liberty configuration is hidden behind the CLI pod.

I think it's best if we call the new attribute `sessionExpireSeconds` and you can do the same thing you can do in Liberty - enter the duration with a hms on the end.  We'll do some preliminary validation using Dan's regex, and then pass it thru to the Liberty configuration which will ultimately parse it and convert it to seconds if necessary.

dacleyra:
as per discussion with @kaczyns , use unit naming consistent with jwtBuilder expiresInSeconds property

https://openliberty.io/docs/ref/config/jwtBuilder.html


@davco01a 

Should be able to consume the envVar by setting server.xml

<jwtBuilder expiresInSeconds=${JwtExpiration} ... >

https://openliberty.io/docs/ref/config/serverConfiguration.html

davco01a:
perfect

PR #217, opened 2019-10-02
Add UI pod and route status to KAppNav status

### LINKED ISSUE #214: Improve KAppNav status reporting
Currently the status reporting for KAppNav just checks to see if it could find the default KAppNav instance.  Since KAppNav is an optional component, the operator can't detect whether it's required or not.  However it should make a better attempt to:
1) determine the version of kappnav (at least report the tag of the container in the pod)
2) determine that it's `ready` by looking at the pod's status and/or conditions
3) go find the route and get the URL that should be used

kaczyns:
Fixes #214.  We're basically trying to improve the usability of the status info.
- If we did not find the default instance of the KAppNav CR in the `kappnav` namespace, we don't report any status for KAppNav.  KAppNav is optional.
- If the UI pod did not start, we report an error and we retry checking the status.
- If the Routes are not accepted (UI and API), we report an error and we retry checking the status.


pkg/controller/kabaneroplatform/kappnav.go
```diff
@@ -34,17 +41,119 @@ func getKappnavStatus(k *kabanerov1alpha1.Kabanero, c client.Client) (bool, erro
...
-	}
 
-	if errors.IsNotFound(err) {
-		k.Status.Kappnav.ErrorMessage = "The default deployment of KAppNav was not found.  KAppNav is an optional component."
+	// If we could not retrieve the default instance, lets assume that
+	// KAppNav is not configured.  We do not want to put any KAppNav status
+	// in the Kabanero status for fear of misleading the user into thinking
+	// that there is a configuration problem.
+	if err != nil {
+		if errors.IsNotFound(err) {
```
mezarin: The If statement does not seem to be necessary here.

PR #221, opened 2019-10-03
Upgrade collection instance when collectionhub url changes

### LINKED ISSUE #220: Collection upgrade does not automatically activate new versions of an existing collection
Currently when the collection URL changes (for example:)
old:
```yaml
spec:
  collections:
    repositories:
    - activateDefaultCollections: true
      name: incubator
      url: https://github.com/kabanero-io/collections/releases/download/v0.1.2/kabanero-index.yaml
```
new:
```yaml
spec:
  collections:
    repositories:
    - activateDefaultCollections: true
      name: incubator
      url: https://github.com/kabanero-io/collections/releases/download/v0.2.0-beta2/kabanero-index.yaml
```
the kabanero controller will iterate the collections in the index to make sure they exist.  However it does not make sure that the versions are correct.  The current behavior is more or less summarized in this PR:
https://github.com/kabanero-io/kabanero-operator/pull/93

I spoke with @stephenkinder today and we agreed that this behavior is not appropriate at this point.  If the collection URL changes, the updated collections should be loaded.  This means the kabanero controller should update the version in the `kind: Collection` when it detects a new one.  The collection controller will then process the upgrade immediately.  

The goal (not yet, but eventaully) will be to allow both versions of the collection to be active at the same time.  For now, we need to be sure to update the version in the collection when we detect that it has changed.

kaczyns:
Fixes #220 
When the collection repository URL changes, presumably with newer or older versions of the same collections that were in the previous URL, the active versions of those collections should reflect what is in the new/updated URL.

If upgrading from the default collection repository 0.1.2 to 0.2.0, the collections should be at the 0.2.0 level.  When downgrading from 0.2.0 to 0.1.2, the collections should be at the 0.1.2 level.

We are just talking about the version of the collections in the collection repository here.  It's valid to have a Kabanero v0.1.2 using the default collection repository 0.2.0, or vice versa.  Upgrade Kabanero should have no impact on the version of the collections.  We are talking about the collection repository URL here.


pkg/controller/kabaneroplatform/featured_collections.go
```diff
@@ -116,7 +116,9 @@ func reconcileFeaturedCollectionsV2(ctx context.Context, k *kabanerov1alpha1.Kab
 
 	ownerIsController := true
 	for _, c := range featured {
-		// For each collection, assure that a corresponding resource exists.
+		// For each collection, assure that a corresponding resource exists and it is at
+		// the highest level found among the repositories.
+		updateCollection := cl.Update
```
mezarin: Shouldn't the same update be needed for V1 collections (method above)? We support both V1 and V2.

kaczyns: Yes but I'd like to abandon the V1 path.

kaczyns: I opened #224 to propose removing the V1 collection path.

PR #255, opened 2019-10-23
Update the Kabanero CSV for modern times

kaczyns:
Changes to the CSV:
- Explicitly code cluster-admin.  Be honest about what we're doing.
- Update the `alm-examples` so they might actually work out of the box today
- Update list of operators and link to soon-to-be-released install-script in the readme
- Update maintainer list
- Update openshift console "hints" for things like versions and status
- Remove required CRDs since that doesn't work anyway
- Remove cluster-mode install support until we decide that's a good idea.

Changes to the Makefile:
- I was copying the wrong CRDs - we are copying the right ones now
- Create kab namespace on push-image in case using the internal registry


kaczyns:
> Hi Tim, i tested this update using operator-courier and got this:
> $operator-courier push ...
> ERROR: crd spec.version not defined. [0.3.0/kabanero_kabanero_crd.yaml]
> ERROR: crd spec.version not defined. [0.3.0/kabanero_collection_crd.yaml]
> Resulting bundle is invalid, input yaml is improperly defined.
> 
> ... if i add the version, then things work again.

Good call - I read up on this:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions
It sounds like the `version` field is deprecated but I see no harm in setting it.

PR #269, opened 2019-10-29
Kabanero modifications to the OCP 4.2 web console

### LINKED ISSUE #228: Can't deploy landing page on OCP 4.2
```
"Error deploying the kabanero landing page.","Request.Namespace":"kabanero","Request.Name":"kabanero","error":"configmaps \"webconsole-config\" not found"
```
Technically the pod is deployed fine, but the next step, updating the OKD console, does not work on 4.2.

kaczyns:
Fixes #228 
Added links to the landing page in the app menu, and the help menu.
There is code here to replace the OpenShift logo with the Kabanero logo, but the design team would like to not do that right now.  I left the code there because it was non-trivial.

Note that there are two commits - the first is vendor updates only, you can ignore it if you don't care.


pkg/controller/kabaneroplatform/landing.go
```diff
@@ -24,10 +28,10 @@ import (
 var kllog = rlog.Log.WithName("kabanero-landing")
 
 // Deploys resources and customizes to the Openshift web console.
-func deployLandingPage(k *kabanerov1alpha1.Kabanero, c client.Client) error {
+func deployLandingPage(k *kabanerov1alpha1.Kabanero, c client.Client, reqLogger logr.Logger) error {
```
mezarin: This is one of those parts that has it's own logr to avoid making the trace/failure look like it is coming from the caller as opposed to this part. I find it easier to debug when the traces show, prominently, the part where the failure happened. Having said that, if you don't want to use the existing logr, you might want to delete it in favor of what you are using.

pkg/controller/kabaneroplatform/landing.go
```diff
@@ -204,270 +208,239 @@ func getLandingURL(k *kabanerov1alpha1.Kabanero, c client.Client, config *restcl
...
+	consoleLink, err = getConsoleLink(c, "kabanero-help-menu-guides")
+	if err == nil {
+		err = c.Delete(context.TODO(), consoleLink)
+		if err != nil {
+			reqLogger.Error(err, "Unable to delete ConsoleLink")
 		}
 	}
-	return false
+
+	// Remove web console config map and customization.
```
mezarin: I know this code is a no-op, but Kind of confusing to have this code active when the code that adds it is commented out. Might be better to comment it out. The call to get will get executed needlessly

pkg/controller/kabaneroplatform/landing.go
kaczyns: I did not realize that this part had its own logger, I should have looked more closely.  I will say though that the logger that I am passing in is modified to show the current request:
```go
reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
I'm not opposed to using a different logger, but I think the team should probably talk about how logging is going to work, as I see the benefit of both modifying the logger to display the current request, but also having the logger indicate which part it's in.

pkg/controller/kabaneroplatform/landing.go
kaczyns: yes I agree, I'll comment it out.

Issue #237, opened 2019-10-14
Operator install to include kabanero-webhook


mchenggit:
We are adding a new container kabanero-webhook, and it needs to be installed with kabanero operator.

kaczyns:
@mchenggit do you need a service/route for your new container?  If so, is it HTTP or HTTPS, and do you need it to have a specific path component (after the hostname)?

mchenggit:
It will be https.

If you can route to us without path component, then we don't need it. Otherwise, please add "/webhook".   The external URL will be:   https://<host>:<port>/webhook/<rest of path>

kaczyns:
I've created a kabanero-operator image that supports what you sent.  You'll need to try it out and let me know if it's close to what you need.

1) Install kabanero-foundation, any version
2) `oc set image -n kabanero deployment/kabanero-operator kaczyns/kabanero-operator:webhook-phase0`
3) Edit your Kabanero CR instance to be `version: 0.3.0` and `Spec.Webhook.Enabled: true`.

Example Kabanero CR instance:
```yaml
apiVersion: kabanero.io/v1alpha1
kind: Kabanero
metadata:
  name: kabanero
  namespace: kabanero
spec:
  collections:
    repositories:
      - activateDefaultCollections: true
        name: incubator
        url: >-
          https://github.com/kabanero-io/collections/releases/download/v0.1.2/kabanero-index.yaml
  webhook:
    enable: true
  version: 0.3.0
```

Let me know when you've had a chance to try this.  The branch where I've made these changes is here:
https://github.com/kaczyns/kabanero-operator/tree/webhook

marikaj123:
Is the work for this issue continue to be targeted for v0.3.0 release?

hibell:
@kaczyns We have some proposed changes to the kabanero-webhook.yaml to be able to set up the TLS listener. It can be found [here](https://github.com/kabanero-io/kabanero-webhook/pull/21). Let me know if you need help with anything. :)

kaczyns:
@marikaj123 Yes this is targeted for v0.3.0.

PR #271, opened 2019-10-30
kabanero-webhook phase 0

kaczyns:
Fixes: #237 
This is "phase 0" of the kabanero-webhook.


deploy/crds/kabanero_v1alpha1_kabanero_crd.yaml
```diff
@@ -122,6 +122,19 @@ spec:
               type: object
             version:
               type: string
+            webhook:
```
mezarin: These same updates need to be made to registry/manifests/kabanero-operator/0.3.0/kabanero_kabanero_crd.yaml

kaczyns: Thanks for catching this.  My intention in the Makefile was to copy the current version of the CRD to the `registery/manifests/kabaner-operator/currentver/` directory, and then when that version is "final", copy the CRD over and move the current version to the next version.  The build does copy the file, but it looks like I also checked in that file at some point.  I will update it, but in the future this should not be necessary.

PR #289, opened 2019-11-08
Add network policy for kabanero and tekton-pipelines namespaces

### LINKED ISSUE #265: Ingress to landing page and CLI pod is broken on OCP 4.2
Ingress to landing page and CLI pod is broken on OCP 4.2

kaczyns:
Fixes #265 
The routes to the Kabanero CLI and Landing page, as well as the Tekton dashboard, do not work reliably.  OpenShift will sometimes report that the pods are not running or the network is not accessible.  

The OpenShift Service Mesh Member Role is watching the kabanero and tekton-pipelines namespaces.  It's watching kabanero so that appsody applications deployed there with serverless capabilities will function, and watching tekton-pipelines so that the github webhook sink will function.  The service mesh assumes that if a namespace is in the member role, that all traffic in and out will come thru istio and therefore traditional routes do not work reliably.

The solution was to add a network policy that allows all traffic to these namespaces.  This configuration is similar to what Kabanero had been using in OKD 3.11 and Istio 1.1.7.  

I've also changed the subscription for the Serverless operator to use the redhat-catalog, since version 1.1.0 is available there now.


kaczyns:
Fixes kabanero-io/kabanero-foundation#126

PR #298, opened 2019-11-14
WIP: Create a pipelines service account during kab install

kvijai82:


deploy/install.sh
```diff
@@ -163,6 +163,9 @@ then
   oc apply -f https://raw.githubusercontent.com/kabanero-io/kabanero-operator/${RELEASE}/deploy/optional.yaml --selector=kabanero.io/component=kappnav
 fi
 
+# Create service account to used by pipelines
+oc apply -f $KABANERO_CUSTOMRESOURCES_YAML --selector kabanero.io/install=24-pipelines-sa
```
kaczyns: You should probably also update the uninstall script to remove these resources.

PR #304, opened 2019-11-15
Use etags to cache remote collection resources

### LINKED ISSUE #147: Update reconciler backoff and/or apply logic for cases when collection apply fails
When a collection apply fails, the collection reconciler will schedule a retry for 60 seconds.  On the retry the reconciler will fetch the collection again from the remote repository.  In some environments this may result in the requestor being throttled.  We should consider an alternative approach.  Some options:
1) Cache the content somewhere (in-memory FS, volume, etc) and only update it if the `Collection` document changes.  The FS should be indexed by the digest or some other way that keeps the versions separate.
2) An exponential backoff on the retry.  This would reduce fetches from the remote repository.
3) Something else.  :-)

kaczyns:
Fixes #147 
The collection controller will get the collection index and the pipeline zip for each collection.  The collection index is always going to be the same, and the pipeline zip is usually the same.  If there is an apply failure, this will happen every 60 seconds... if all of the collections have an apply failure this can happen a lot.  Eventually GitHub will throttle the cluster.  This PR caches the data using the `etag` that GitHub returns in the HTTP headers on each response.

There is one cache for the process.  Every 30 minutes, a task runs to purge the cache.  Any entry nt referenced in the past 12 hours will be purged.

There is no attempt to make this a high-performing cache.  There is a mutex that synchronizes access to the backing map, which is keyed by URL.  The mutex is not held around the HTTP request itself, to avoid holding the lock for an excessive period of time.  The trade-off is that the cache can change while the request is running, but there is not a lot of impact if the cache is not correct (it's possible two threads will add the same data after retrieving it from the host).

The test cases test for:
1) Data can be cached
2) Data can be changed (when GitHub sends a new etag)
3) Data can be read when there is no etag
4) Data can be purged from the cache

### DIFF
--- a/pkg/controller/collection/archive.go
+++ b/pkg/controller/collection/archive.go
@@ -9,8 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
-	"net/http"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


jgawor:
What happens if etags are not there?

kaczyns:
> What happens if etags are not there?

Nothing - we don't cache the data.

Issue #308, opened 2019-11-19
Add digest as a directive on the collection directive processor


kaczyns:
In support of #288 
Collections can use directives in the first line of an asset to instruct the collection controller to make a substitution. For example:
```
#Kabanero! on activate substitute CollectionId for text 'CollectionId'
```
Will replace the text `CollectionId` in the yaml, with the ID of the collection that is being applied.  Here is a collection asset that uses this:
https://github.com/kabanero-io/kabanero-pipelines/blob/bf98716083a5a66fa03aa7372015140ec00be0d5/pipelines/incubator/deploy-task.yaml

Currently, `CollectionName` and `CollectionId` are supported, although CollectionName is of little use in Kubernetes since it can contain unsupported characters.  We should add a third supported value, `Digest`.  The digest will resolve to the first 8 characters of the SHA256 for the pipeline zip containing the asset.  If a pipeline zip has a SHA256 of `abcdef0123456789` and the following asset exists in the zip:
```yaml
#Kabanero! on activate substitute Digest for text 'Digest'
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: deploy-task-Digest
spec:
  inputs:
    resources:
    - name: git-source
...
```
then Kabanero would create a `kind: Task` with name `deploy-task-abcdef01`.  And so on for each asset in the collection.  Each asset in the collection (pipeline zip) would use the same digest since they all come from the same zip.

Method `process_directive` in `directive_processor.go` takes a context map as input, and the first 8 characters of the SHA256 for the pipeline zip will need to be added to the map, using `Digest` as the key, so that the processor can make the substitution.

PR #345, opened 2019-12-05
Share assets among different collections

### LINKED ISSUE #309: Allow multiple collections to use the same pipeline assets
In support of #288 
If all of the collections in a collection hub use the same pipeline zip, it would be good if they could all use the same instances of the assets (same Kubernetes object instance).  Currently when the collection controller creates assets, it sets the owner to the collection that created the asset.  If we are going to share assets among collections, we should maintain a list of owners in the asset as opposed to a single owner.  This is supported by Kubernetes.

This will change the way we create and delete assets.  If a collection thinks it needs to create an asset, it should first try to `get` the asset.  If it's already there, then the collection should add itself as an owner.  Similarly, if a collection thinks it needs to delete an asset, it should remove itself from the owner list.  If the owner list is not empty, the asset should not be deleted.  If the owner list is empty (or if this collection is the only owner) then the asset can be deleted.

When a collection changes versions, it is probably a good idea to check to see if the assets the new version is going to apply, are the same as what will be deleted, in which case the whole thing is a no-op.  The SHA256 of the pipeline zip could be used to determine this, or some other way.  This is of course optional, as there would be no harm in deleting, and then re-creating the assets.

kaczyns:
Fixes #309 
This is a small step on the journey to having multiple versions of a collection active at the same time.  I've re-written the activate and deactivate methods on the collection controller, to compare the desired version in the `Spec` against the active version(s) in the `Status`.  Both the version of the collection, and the SHA of the pipeline zip, are used to determine whether the collection assets need to be refreshed or not.  Ideally the assets in the pipeline zip contain the SHA in the asset name (what was delivered in #308) but this is not strictly required.  If the assets exist already, we use the owner references array in the `ObjectMeta` to keep track of the additional collections that are using the asset.  When a collection is no longer using the asset, it will remove itself from the owners, or if it's the last owner, it will be deleted.  If the collection is deleted, Kube object deletion will remove the collection from the owner list of all the assets, removing the asset if it's the last owner.

A couple of un-related fixes are in here as well.  I added an `xxd`-style trace function, which was useful in debugging checksum problems.  And I added the accepted encodings header on the archive get requests to prevent sites other than github.com from removing the gzip encoding on the pipeline zip (the checksum doesn't match when you get a tar but you wanted a tar.gz.  It also doesn't uncompress very well.)

There are quite a lot of test cases for the collection controller now.  The new test cases use the `httptest` package to serve up pipeline zips.  I added a mock kube client which only keeps track of object names and owner references, to make sure that the owner reference logic works OK when sharing assets between the collections.


kaczyns:
I updated this PR with some additional code to support activating multiple versions of the same collection.  Right now, since the CRD doesn't support this, the unit tests are the only code exploiting the multi-version path.  There are quite a few unit tests.

Issue #311, opened 2019-11-20
Update status with error  when a new version of collection is not activated due to error condition


mtamboli:
I updated Kabanero CR to a new release of collection. It seems like that one of the collections has some errors in the customization of pipelines. So all of the collections were activated to new release except the one which had issues. This behavior is correct. But it would  be nice to display the error in collection status or some other place which indicates there is a problem with new collections.

**Collection which had issue activating new version:**
```
spec:
  desiredState: active
  name: java-microprofile
  repositoryUrl: >-
    https://github.com/mtamboli/collections/releases/download/v0.3.6/kabanero-index.yaml
  version: 0.2.19
status:
  activeLocation: >-
    https://github.com/mtamboli/collections/releases/download/v0.3.5/kabanero-index.yaml
  activePipelines:
```
**Collection which got activated to the new version correctly:**
```
spec:
  desiredState: active
  name: java-spring-boot2
  repositoryUrl: >-
    https://github.com/mtamboli/collections/releases/download/v0.3.6/kabanero-index.yaml
  version: 0.3.16
status:
  activeLocation: >-
    https://github.com/mtamboli/collections/releases/download/v0.3.6/kabanero-index.yaml
  activePipelines:
```
Had to look at the operator log to know there was an error:
```
Header names:  ./,./build-push-deploy-pipeline.yaml,./._deploy-task.yaml,./deploy-task.yaml,./image-retag-push-task.yaml,./manifest.yaml,./._test-task.yaml,./test-task.yaml,./._image-scan-task.yaml,./image-scan-task.yaml,./._test-build-push-deploy-pipeline.yaml,./test-build-push-deploy-pipeline.yaml,./._build-push-task.yaml,./build-push-task.yaml
[...]

PR #354, opened 2019-12-10
collection status err

dacleyra:
Makefile
split install from CR apply, quicker to test operator change
fix sed for collection controller

Collection Controller
move errorMessage up in scope
GetManifests 1st instance, return err instead of nil
GetManifests 2nd instance, log & update status. This should address #311, as it will appear as below

```
kind: Collection
metadata:
  creationTimestamp: "2019-12-10T17:11:30Z"
  generation: 1
  name: java-microprofile
  namespace: kabanero
  ownerReferences:
  - apiVersion: kabanero.io/v1alpha1
    controller: true
    kind: Kabanero
    name: kabanero
    uid: 5657f16b-1a99-11ea-a07f-00000a100cb6
  resourceVersion: "20517833"
  selfLink: /apis/kabanero.io/v1alpha1/namespaces/kabanero/collections/java-microprofile
  uid: 1b8dd1ce-1b70-11ea-a07f-00000a100cb6
spec:
  desiredState: active
  name: java-microprofile
  repositoryUrl: https://github.com/mtamboli/collections/releases/download/v0.3.6/kabanero-index.yaml
  version: 0.2.19
status:
  statusMessage: 'Error during version change from  to 0.2.19: File ./._deploy-task.yaml
    was found in the archive, but not in the manifest.yaml'
```

@mtamboli

### DIFF
--- a/pkg/controller/collection/collection_controller.go
+++ b/pkg/controller/collection/collection_controller.go
@@ -539,6 +539,8 @@ func activate(collectionResource *kabanerov1alpha1.Collection, collection *Colle
 		"CollectionName": collection.Name,
 	}
 
+	errorMessage := "Error during version change from " + collectionResource.Status.ActiveVersion + " to " + collection.Version
+
 	// Detect if the version is changing from the active version.  If it is, we need to clean up the
 	// assets from the previous version.
 	if (collectionResource.Status.ActiveVersion != "") && (collectionResource.Status.ActiveVersion != collection.Version) {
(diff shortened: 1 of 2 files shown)


pkg/controller/collection/collection_controller.go
```diff
@@ -613,6 +614,8 @@ func activate(collectionResource *kabanerov1alpha1.Collection, collection *Colle
 		// Retrieve manifests as unstructured
 		manifests, err := GetManifests(pipeline.Url, pipeline.Sha256, renderingContext, log)
 		if err != nil {
+			log.Error(err, errorMessage, "resource", pipeline.Url)
+			collectionResource.Status.StatusMessage = errorMessage + ": " + err.Error()
```
kaczyns: I think we should use a different error message here - if there was no previous version, it's not an upgrade, just an activate.  Maybe leave the old errorMessage where it was, and put a new one here inside this scope?

dacleyra: ok, updated as such

PR #366, opened 2019-12-17
Add OAuth information in landing / console pod

### LINKED ISSUE #355: Add GitHub OAuth variables to Kabanero CRD spec
The console (previously called "landing") application is going to start using GitHub OAuth services to get information from the logged in user. This is the start of our role based approach.

The console app will need some information about OAuth such as:

(required)
* OAuth client ID - obtained from creating an OAuth app in GitHub settings
* OAuth client secret - obtained from creating an OAuth app in GitHub settings
   * Maybe these should be in a kubernetes secret instead of the CRD

(Optional, github.com is default)
* authorizationEndpoint - for example: `https://github.com/login/oauth/authorize`
* tokenEndpoint - for example: `https://github.com/login/oauth/access_token`
* userApi - for example: `https://api.github.com/user/emails`
* website - for example: `https://github.com`

these are all configurable options from the [Liberty GitHub Social Login feature](https://www.ibm.com/support/knowledgecenter/en/SSEQTP_liberty/com.ibm.websphere.liberty.autogen.nd.doc/ae/rwlp_config_githubLogin.html). I believe the listed are the only ones really needed depending on where the GitHub server is

If using Public GitHub.com, then only `client ID` and `client secret` of these need to be specified, as social login uses github.com as the default configured

I was thinking a good spot for these might be in the Kabanero `spec.github`. One question though, is there a way to inject these values as env vars to the console pods? Our liberty app has a server xml feature configuration (social login) that would need to read these values on start up.

kaczyns:
Fixes #355 
I added the OAuth configuration requested in #355.  Most of this is in the deployment object for the landing page.  There had been some resistance to creating deployments programmatically instead of using the orchestration (yaml) with transformations.  When I added the OAuth information to the landing page deployment, I converted both the landing page deployment and the CLI deployment to use the orchestrations.  There are no more programmatic deployments in the code.

I had to add two new transformations for OAuth - one for adding an environment variable, and one for mounting a secret as a volume.


pkg/controller/kabaneroplatform/landing.go
```diff
@@ -70,27 +71,79 @@ func deployLandingPage(k *kabanerov1alpha1.Kabanero, c client.Client) error {
...
+		Name:      secretName,
+		Namespace: k.ObjectMeta.Namespace}, secretInstance)
+
+	if err == nil {
+		// The secret exists, go ahead and put in the oauth information
+		transforms = append(transforms, kabTransforms.MountSecret(secretName, "/etc/oauth"))
+
+		// Try and figure out what the OAuth URL for github is, based on the API url.  If the API
+		// URL is not set, assume that it is "api.github.com".
+		apiUrlString := k.Spec.Github.ApiUrl
```
alohr51: Are the properties in `k.Spec.Github.ApiUrl` case insensitive? Our docs show `spec` `github` and `apiurl` as all lowercase when configuring a CR instance.

pkg/controller/kabaneroplatform/landing.go
```diff
@@ -70,27 +71,79 @@ func deployLandingPage(k *kabanerov1alpha1.Kabanero, c client.Client) error {
...
+	// See if we should define the OAuth volume and variables
+	secretInstance := &corev1.Secret{}
+	secretName := "kabanero-github-oauth-secret"
+	err = c.Get(context.Background(), types.NamespacedName{
+		Name:      secretName,
+		Namespace: k.ObjectMeta.Namespace}, secretInstance)
+
+	if err == nil {
+		// The secret exists, go ahead and put in the oauth information
+		transforms = append(transforms, kabTransforms.MountSecret(secretName, "/etc/oauth"))
```
alohr51: so for the scenario when a user creates the secret named `kabanero-github-oauth-secret` and adds two key value pairs (client ID, and secret). Do we need to tell them to name the keys x and y something specific?

The way I assume the secret is mounted to `/etc/oauth` is - it will create two files in the container at `/etc/oauth`, and the filenames are named after the keys given when creating the secret. I'll have to know those filenames for my automation to grab the right values out.

This is more a doc question if we will have to document to tell them name the keys `githubClientID` and `githubClientSecret` or something.

alohr51:
@kaczyns I looked at the kab landing/console pieces and reviewed most of what I could. Thanks for getting this out so quick!

pkg/controller/kabaneroplatform/landing.go
kaczyns: The golang structs for the mapping of the CRD always uppercase the first letter of each property, because that's how go differentiates public and private things.  If the first letter is upper case, it's public, and if not, it's private.

pkg/controller/kabaneroplatform/landing.go
kaczyns: Yes, I assumed that there would be doc that said something like "Create a secret named `kabanero-github-oauth-secret` and add two keys/values.  You will need to pick the keys and then the console will need to look for those in `/etc/oauth`.

pkg/controller/transforms/secret.go
```diff
@@ -0,0 +1,108 @@
...
+			var newVolumeMounts []interface{}
+			if (err == nil) && (ok) {
+				// Look and see if this volume mount exists already
+				for _, volumeMountRaw := range volumeMounts {
+					volumeMount, ok := volumeMountRaw.(map[string]interface{})
+					if !ok {
+						return fmt.Errorf("Could not assert map type for volume mount: %v", volumeMountRaw)
+					}
+
+					// Copy all the volume mounts to the new list, skipping the desired name if it exists.
```
mezarin: Same comment as previous one.

pkg/controller/transforms/secret.go
```diff
@@ -0,0 +1,108 @@
...
+		volumes, ok, err := unstructured.NestedSlice(u.Object, "spec", "template", "spec", "volumes")
+		var newVolumes []interface{}
+		if (err == nil) && (ok) {
+			for _, volumeRaw := range volumes {
+				volume, ok := volumeRaw.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("Could not assert map type for volume: %v", volumeRaw)
+				}
+
+				// Copy all the volumes to the new list, skipping the desired name if it exists.
```
mezarin: Same comment as previous two.

Issue #250, opened 2019-10-22
Update to operator-sdk v0.11.0 (or newer)


dacleyra:
Require controller-runtime v0.2.0 or newer to support issue #230

kaczyns:
Some more words about this....
The operator-sdk provides the initial dependencies for your project (Gopkg.toml) which include controller-runtime.  The version of operator-sdk that the kabanero-operator was created with, used controller-runtime v0.1.10.  The v0.1 train stops at v0.1.12 (or at least that's where it's stopped right now).

We know of at least one bug that we're currently working around, plus the stuff for #230, that would be fixed by moving up to controller-runtime v0.2.0.  We can't just upgrade controller-runtime by itself, because then anything we do with `operator-sdk` will not work (ie add a new CRD and controller).  The two should be updated together.  Or, we need to stop using `operator-sdk`, and I don't think we're ready to do that yet.

The operator-sdk 0.11.0 includes controller-runtime v0.2.0.  The community has acknowledged that this is a non-trivial migration but one that is going to be necessary.

kaczyns:
This will be included in Kabanero 0.5.0.

PR #369, opened 2019-12-31
Upgrade to operator-sdk 0.11.0 and controller-runtime 0.2.0

kaczyns:
This was more work than I anticipated.  I stepped thru the operator-sdk versions one by one.  The first commit gets us to 0.10.0 which was fairly painless.  The second commit gets us to 0.11.0.  If you're only going to review one of these, review the second one, as that's the more interesting (and also has fewer files to look at, believe it or not).  For the most part I followed the operator-sdk [upgrade guide](https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md) and controller-runtime release notes.

First commit:
* Upgrade to operator-sdk 0.10.0
* Switch to go modules.  Dependencies are now managed as modules.  There is no more vendor directory.  You can find the modules in `$GOPATH/pkg/mods` now.  The `Makefile` now prefixes `go` commands with `GO111MODULE=on` to force module support.

Second commit:
* Upgrade to operator-sdk 0.11.0 and controller-runtime 0.2.0
* controller-runtime webhooks no longer generate their own certificates.  Switch to using the OpenShift service certificates.  Reading the CA certificate and injecting it into the webhook configuration was fun.
* operator-sdk and controller-tools have different ideas about what the plural form of `kabanero` should be, so there are some additional markers (annotations) in the `kabanero_types.go` file to compensate for this.
* The CRD file name emitted from operator-sdk changed
* Lots of little changes where controller-runtime package names and type names changed


kaczyns:
Fixes #250

PR #372, opened 2020-01-03
fix up description for kabanero manifests

alohr51:
fix the description in 2 ways:

1. fix instal typo
1. change link for location of install script from the Kabanero foundation repo to the operators releases and tell the users to look under "Assets" for each release to find it.


kaczyns:
Thanks for spotting this Andrew!  Merged.

PR #375, opened 2020-01-09
Fix pipelinerun tests

jgawor:
Wrong `APP` name was used when validating AppsodyApplication and KnativeService.


kaczyns:
@dacleyra would you mind reviewing this please?

PR #377, opened 2020-01-10
Validate collection Id.

### LINKED ISSUE #167: validate CollectionID so it doesn't break Pipeline variable substitution
See https://github.com/kabanero-io/kabanero-operator/issues/114#issuecomment-528899461
Since the CollectionId is being used  as a substitution/variable in the Pipelines we need to validate / restrict the field so that pipelines don't fail with invalid characters.

mezarin:
Fixes #167

### DIFF
--- a/pkg/controller/collection/collection_controller.go
+++ b/pkg/controller/collection/collection_controller.go
@@ -3,6 +3,7 @@ package collection
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -24,14 +25,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 var log = logf.Log.WithName("controller_collection")
+var cIDRegex = regexp.MustCompile("^[a-z]([a-z0-9-]*[a-z0-9])?$")
 
 const (
 	// Asset status.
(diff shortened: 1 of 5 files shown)


pkg/controller/collection/collection_controller.go
```diff
@@ -465,6 +467,24 @@ func getNamespaceForObject(u *unstructured.Unstructured, defaultNamespace string
...
+	// Gather the known collection asset (*-tasks, *-pipeline) substitution data.
+	renderingContext := make(map[string]interface{})
+	if len(collections) > 0 {
+		// The collection id is the name of the Appsody stack directory ("the stack name from the stack path").
+		// Appsody stack creation namimg constraints:
+		// "The name must start with a lowercase letter, contain only lowercase letters, numbers, or dashes,
+		// and cannot end in a dash."
+		cID := collections[0].collection.Id
+		if !cIDRegex.MatchString(cID) {
+			return fmt.Errorf("collection Activation failed because an invalid collection id of %v was found. The collection id value must follow stack creation name rules", cID)
```
kaczyns: is there someplace or some thing that we can point the user to, to show what the rules are?  I don't know if Appsody has a web page or something we can link to.

pkg/controller/collection/collection_controller_test.go
```diff
@@ -337,60 +341,166 @@ func (ch collectionHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
 
 var triggerPipeline = fileInfo{
-	name: "/trigger.pipeline.tar.gz",
+	name:   "/trigger.pipeline.tar.gz",
 	sha256: "bba27927b6da1323f264e957c526c4df118d657dd8ad19e08994d70caaf31f73"}
 
+// --------------------------------------------------------------------------------------------------
+// Test collection/stack id validation.
+// --------------------------------------------------------------------------------------------------
+func TesCollectionIDValidation(t *testing.T) {
```
kaczyns: Typo - Tes -> Test, these tests won't run if the method name is spelled this way.

pkg/controller/collection/collection_controller.go
```diff
@@ -24,14 +25,15 @@ import (
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 var log = logf.Log.WithName("controller_collection")
+var cIDRegex = regexp.MustCompile("^[a-z]+[a-z0-9-]*[a-z0-9]+$")
```
kaczyns: With this regex, a collection ID of `a` will fail.  I'm not sure how to write a regex that will catch this case... maybe if the length of the ID is 1, we can just check to see if it's a valid lower case letter?

mezarin: Thanks for catching this Tim. Updated a new commit.

PR #392, opened 2020-01-15
Introduce v1alpha2 kabanero and stack crd

### LINKED ISSUE #386: Add v1alpha2 and Stack CRDs
We have received feedback from users that the idea of an Appsody stack and a Kabanero collection being the same thing, is confusing.  A brief overview - a Kabanero collection starts as an Appsody stack, to which we add additional configuration for Kabanero such as pipelines, triggers and image names.  The app developer generally uses Appsody (and an appsody stack) to develop an application, while operations is using Kabanero (and Kabanero collections) to deploy the applications to test and production.

The Kabanero controller should be taught how to read an Appsody stack, with the additional bits and pieces required by Kabanero specified in the Kabanero CR instance, rather than in the Appsody stack.  To this end, we should work towards a v1alpha2 of the Kabanero CRD.  The Collection CRD should be re-named to the Stack CRD to better reflect what it will actually be.

See:
https://github.com/kabanero-io/roadmap/issues/126

kaczyns:
Resolves #386 
This is the proposed v1alpha2 CRD for Kabanero and Stacks.  The main change here as compared with v1alpha1 is that the repositories listed in the Kabanero CR instance are Appsody repositories and not Collection Hub repositories.  Since Appsody repositories don't have pipeline or trigger information, that information is now supplied in the v1alpha2 Kabanero CR instance, and then propagated to the v1alpha2 Stack CR instance.  The Stack CR instance is roughly equivalent to what a v1alpha1 Collection CR instance was.


kaczyns:
Closing this in favor of #408

PR #395, opened 2020-01-16
add keycloak

meiaus:
modified:   deploy/install.sh
modified:   deploy/uninstall.sh
new file:   deploy/keycloak-install.sh
new file:   deploy/keycloak-uninstall.sh

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -225,18 +225,18 @@ then
   oc apply -f https://raw.githubusercontent.com/kabanero-io/kabanero-operator/${RELEASE}/deploy/optional.yaml --selector=kabanero.io/component=kappnav
 fi
 
-# Install Keycloak if selected
-if [ "$ENABLE_KEYCLOAK" == "yes" ]
-then
-   curl -L -s https://github.com/kabanero-io/kabanero-operator/releases/download/$RELEASE/keycloak-install.sh | bash 
-fi
-
 # Create service account to used by pipelines
 oc apply -f $KABANERO_CUSTOMRESOURCES_YAML --selector kabanero.io/install=24-pipeline-sa
 
 # Role used by the collection controller to manipulate triggers in the
 # tekton-pipelines namespace (for use by tekton github webhooks extension)
 oc apply -f $KABANERO_CUSTOMRESOURCES_YAML --selector kabanero.io/install=25-triggers-role
+
+# Install Keycloak if selected
+if [ "$ENABLE_KEYCLOAK" == "yes" ]
+then
+   curl -L -s https://github.com/kabanero-io/kabanero-operator/releases/download/$RELEASE/keycloak-install.sh | bash 
+fi
 
 # Install complete.  give instructions for how to create an instance.
 SAMPLE_KAB_INSTANCE_URL=https://raw.githubusercontent.com/kabanero-io/kabanero-operator/${RELEASE}/config/samples/default.yaml
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -10,6 +10,7 @@ SLEEP_SHORT="${SLEEP_SHORT:-2}"
 
 # Optional components (yes/no)
 ENABLE_KAPPNAV="${ENABLE_KAPPNAV:-no}"
+ENABLE_KEYCLOAK="${ENABLE_KEYCLOAK:-no}"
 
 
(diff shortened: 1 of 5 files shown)


deploy/install.sh
```diff
@@ -224,6 +225,15 @@ then
   oc apply -f https://raw.githubusercontent.com/kabanero-io/kabanero-operator/${RELEASE}/deploy/optional.yaml --selector=kabanero.io/component=kappnav
 fi
 
+# Install Keycloak if selected
+if [ "$ENABLE_KEYCLOAK" == "yes" ]
+then
+   #final official URL but is not tested yet
+   # bash <(curl -s https://raw.githubusercontent.com/kabanero-io/kabanero-operator/${RELEASE}/deploy/keycloak-install.sh)
+   # For testing purpose: temporarily using GSA server before final URL available
+    bash <(curl -s -u meichang:mypwd https://rtpgsa.ibm.com/projects/w/wassvt/PersonaTest/meichang/keycloak-install.sh)
```
kaczyns: This URL can't be used as no one outside of IBM will be able to access it.  You need to host this on a public branch in github, or somewhere else.

deploy/keycloak-install.sh
```diff
@@ -0,0 +1,22 @@
...
+
+ssotemplate=$(oc get templates -n openshift -o=custom-columns=NAME:.metadata.name --ignore-not-found | grep sso73-x509-postgresql-persistent)
+
+if [[ ${ssotemplate} ]]; then
+   echo "Installing Keycloak ..."
+   echo "Copying the template ${ssotemplate} to json file ..."
+   oc get template ${ssotemplate} -n openshift -o json > ${ssotemplate}.json
+   echo "The template json file is: " ${ssotemplate}.json
+   echo "Modifying the josn file with postgresql version 9.6 and mountpath to resolve known issues on postgresql ..."
+   sed -i 's/"9.5"/"9.6"/g' ${ssotemplate}.json
```
kaczyns: How confident are you that this template will always be present, and that it won't change in such a way that will make these `sed` commands break?  I will be honest and say that I am not very familiar with OpenShift templates and what their lifecycle is.  Is this likely to work in OCP 4.3?  

I am wondering if a better approach would be to retrieve the JSON once, modify it as needed, and then check in those modifications into our `optional.yaml` file.

meiaus:
Quick update after 0116 review and per slack discussion:

1. update .travis.yml to add files for deploy/keycloak-install.sh and deploy/keycloak-uninstall.sh
2. update install.sh and uninstall.sh to use release URL location for keycloak-install.sh and keycloak-install.sh when running bash curl.
3. update keycloak-install.sh to add checking for only installed once.

@brutif  checking: OCP4.3 template catalog if same template exists

meiaus:
0117 update:
1. update keycloak-install.sh to add conditions check before 'sed' update
2. update deploy/uninstall.sh to move the keyloak deletion lines to near the bottom before ```### Subscriptions```.

Per @brutif, OCP 4.3 contains the same template as in OCP 4.2.

@kaczyns question: Please advise how to test from the repo once no more update is needed at this time. Thanks.

deploy/install.sh
meiaus: fixed

deploy/keycloak-install.sh
meiaus: fixed

meiaus:
> 
> 
> `sed -i` without a backup file pattern is invalid on MacOS.
> 
> https://unix.stackexchange.com/a/381201/89196

Thanks @covener for the reminder/suggestion.  The script is updated accordingly.

meiaus:
0118 update:
1. deploy/keycloak-install.sh and deploy/keycloak-uninstall.sh to add `set -Eeox pipefail`
2. deploy/keycloak-install.sh to add `sed -i.bak`, add if-condition on `grep` to accommodate the usage of `set -Eeox pipefail` and remove json and .bak files
3. deploy/install.sh to move keycloak install further down before `# Install complete`

meiaus:
0121 update:
1. deploy/keycloak-install.sh to change the to-be-updated postgresql version from 9.6 to 10 and add single quotes in `grep` lines

meiaus:
Closing the PR since the update is not required for kabanero this time.  (Working with CP4App install instead.)

PR #408, opened 2020-01-22
388

dacleyra:
merge kaczyns/stackCrd
Add the stack-controller to the Makefile
Copy/Rename collective-controller to stack-controller
Modify tests that relied on the singleton
Modify tests that relied on repositoryUrl
In the controller, build the matchingStacks Stack array of singletons from the StackSpec Versions array (don't rely on parent Kabanero or repo url)


pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+	}
+	return false
+}
+
+
+
+
+// Finds the stack with the highest semver.  Caller has validated that resolvedStack
+// contains at least one stack.
+func findMaxVersionStack(stacks []resolvedStack) *resolvedStack {
```
kaczyns: We can delete this function

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+				maxVersion = curVersion
+			}
+		}
+	}
+
+	// It's possible we didn't find a valid semver, in which case we'd return nil.
+	return maxStack
+}
+
+func (r *ReconcileStack) ensureStackHasOwner(c *kabanerov1alpha2.Stack, k *kabanerov1alpha2.Kabanero) error {
```
kaczyns: Can delete this function, see comment later about why

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Resolve the kabanero instance
+	// TODO: issue #92, when repo is added to the Stack, there will be no need for the Kabanero
+	//       object here.
+	var k *kabanerov1alpha2.Kabanero
```
kaczyns: I don't think it's necessary for us to resolve the kabanero instance anymore, hopefully this block of code can be removed.

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+	beingDeleted, err := processDeletion(ctx, instance, r.client, reqLogger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if beingDeleted {
+		return reconcile.Result{}, nil
+	}
+
+	rr, err := r.ReconcileStack(instance, k)
```
kaczyns: Hopefully do not need to pass kabanero instance to this method anymore, see later comments...

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+	var stackName string
+	if c.Spec.Name != "" {
+		stackName = c.Spec.Name
+	} else {
+		stackName = c.Name
+	}
+
+	r_log = r_log.WithValues("Stack.Name", stackName)
+
+	// A stack created by the CLI might not have kabanero as the owner.
```
kaczyns: Can not assume that a Kabanero instance owns a stack anymore... so can remove this call to ensure owner.

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+	if err != nil {
+		r_log.Error(err, "Could not make kabanero the owner of the stack")
+	}
+
+
+
+	// Retreive all matching stack names from all remote indexes.
+	// TODO: Start using the URL in the Stack object so we don't need to reference the
+	//       parent Kabanero anymore.
+	var matchingStacks []resolvedStack
```
kaczyns: ok this whole block here... we don't need to worry about what indexes are in the kabanero CR instance anymore, nor what version they are.  just always iterate the versions in the Stack CR instance, and build out the collection list manually, like you're doing in the `case "v2":` block below.

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+
+		default:
+			fmt.Sprintf("Index is unsupported version: %s", apiVersion)
+		}
+	}
+
+
+
+
+/* - TODO - Delete, no longer valid?
```
kaczyns: Yes - can delete all of this commented out code

pkg/controller/stack/stack_controller.go
```diff
@@ -0,0 +1,981 @@
...
+			r_log.Error(semverErr, "Could not determine upgrade availability for stack "+stackName)
+		}
+	}
+*/
+
+	// Process the versions array and activate (or deactivate) the desired versions.
+	err = reconcileActiveVersions(c, matchingStacks, r.client)
+
+
+/* - TODO - Delete, no longer valid?
```
kaczyns: correct - no longer valid

PR #410, opened 2020-01-22
go.mod

dacleyra:
Local builds are complaining about the sum
On my local and @stevenschader we both get this sum 
FWIW our operator-sdk is the linux x86 binary


kaczyns:
It appears that go 1.13.x generates a different hash (?) than 1.12.x.  So this all works fine when compiling with go 1.13.  I've updated `.travis.yml` to require go 1.13.x.  Lets hold off on merging this PR until we decide whether we really want to make the jump of compiling the operator with go 1.13.

kaczyns:
FWIW the binary for operator-sdk 0.11.0 was compiled with go 1.12.7.

PR #414, opened 2020-01-23
402

dacleyra:
- Landing: ensure URL.Scheme is at least defaulted to https
- CLI: process as net/url as per Landing
  - Will now set a url if the parse fails, but behavior still does not set if empty

### DIFF
--- a/pkg/controller/kabaneroplatform/cli.go
+++ b/pkg/controller/kabaneroplatform/cli.go
@@ -93,7 +93,7 @@ func reconcileKabaneroCli(ctx context.Context, k *kabanerov1alpha1.Kabanero, cl 
 			reqLogger.Error(err, "Could not parse Github API url %v, assuming api.github.com", apiUrlString)
 			apiUrl, _ = url.Parse("https://api.github.com")
 		}
-		transforms = append(transforms, kabTransforms.AddEnvVariable("github.api.url", apiUrlString))
+		transforms = append(transforms, kabTransforms.AddEnvVariable("github.api.url", apiUrl.String()))
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/kabaneroplatform/cli.go
```diff
@@ -83,7 +84,16 @@ func reconcileKabaneroCli(ctx context.Context, k *kabanerov1alpha1.Kabanero, cl
 
 	// Export the github API URL, if it's set.  This is used by the security portion of the microservice.
 	if len(k.Spec.Github.ApiUrl) > 0 {
-		transforms = append(transforms, kabTransforms.AddEnvVariable("github.api.url", k.Spec.Github.ApiUrl))
+		apiUrlString := k.Spec.Github.ApiUrl
+		apiUrl, err := url.Parse(apiUrlString)
+		if len(apiUrl.Scheme) == 0 {
```
kaczyns: `url.Parse()` returns a pointer to a `URL`, so I don't think it's safe to call `len(apiUrl.Scheme)` here.  I think we need to check for err first, and then try to parse the scheme.

pkg/controller/kabaneroplatform/landing.go
```diff
@@ -120,6 +120,9 @@ func deployLandingPage(k *kabanerov1alpha1.Kabanero, c client.Client) error {
 		}
 
 		apiUrl, err := url.Parse(apiUrlString)
+		if len(apiUrl.Scheme) == 0 {
```
kaczyns: Same comment here, we need to check err first before calling `len(apiUrl.Scheme)`

PR #418, opened 2020-01-24
update pipeline name for int tests

dacleyra:
Merge when 0.5.0-rc.4 collections release is available

https://github.com/kabanero-io/collections/pull/254


kvijai82:
@dacleyra One more thing. We just discovered that in the pipelinerun, `serviceAccount` has now changed to `serviceAccountName` in Tekton 0.8.  Could you also please update that?

dacleyra:
@kvijai82 updated SA

PR #427, opened 2020-01-27
420

dacleyra:
Install
 - Tweak CatalogSource status workaround
 - Remove smcp cr.  Required smcp & smmr crs are created & owned by knative

```
NAMESPACE                 NAME      MEMBERS
knative-serving-ingress   default   [knative-serving]
---
NAMESPACE                 NAME            READY
knative-serving-ingress   basic-install   True
```

Update kabanero registry to include servicemesh 1.0.4, serverless 1.3.0
Update subscriptions to 0.6

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -121,18 +121,27 @@ oc delete -f $KABANERO_SUBSCRIPTIONS_YAML  --ignore-not-found --selector kabaner
 
 ### CatalogSource
 
+# Install Kabanero CatalogSource
+oc apply -f $KABANERO_SUBSCRIPTIONS_YAML --selector kabanero.io/install=00-catalogsource
+
+# Check the kabanero-catalog CatalogSource pod is Running
+unset STATUS
+until [ "$STATUS" == "Running" ]
+do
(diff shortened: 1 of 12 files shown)


deploy/install.sh
```diff
@@ -116,21 +121,27 @@ oc delete -f $KABANERO_SUBSCRIPTIONS_YAML  --ignore-not-found --selector kabaner
...
 
+# Install Kabanero CatalogSource
+oc apply -f $KABANERO_SUBSCRIPTIONS_YAML --selector kabanero.io/install=00-catalogsource
+
+# Check the kabanero-catalog CatalogSource pod is Running
+unset STATUS
+until [ "$STATUS" == "Running" ]
+do
+  echo "Waiting for CatalogSource kabanero-catalog pod to be ready."
+  STATUS=$(oc -n openshift-marketplace get pod -l olm.catalogSource=kabanero-catalog --output=jsonpath={.items[0].status.phase})
```
kaczyns: If we're replacing a previous kabanero catalog, how can we be sure we're not getting the status of the pod that was serving the previous catalog, vs the one we just applied?

Issue #404, opened 2020-01-22
Installation fails on OCP 4.3


csantanapr:
Environment:
- OCP 4.3
- Cloud Pak for Applications 4.0.0 (kabanero operator image `image: kabanero/kabanero-operator:0.3.4`)

Steps to reproduce:
- Run CP4Apps installer that installs kabanero operator

Results:
```
The InstallPlan status is 'Failed'. There might a problem that is preventing the InstallPlan installation from completing, or additional time might be required to complete.


Run the following commands to examine the status of the InstallPlan:
oc -n "kabanero" describe InstallPlan "install-htgqh"


Once the problem is resolved or the status of the InstallPlan becomes 'Complete', re-run the installer to complete the installation process.


oc -n "kabanero" describe InstallPlan "install-htgqh"
```

After running the command to describe the first failed components contains
```
Name:         install-htgqh
Namespace:    kabanero
Labels:       <none>
Annotations:  <none>
API Version:  operators.coreos.com/v1alpha1
Kind:         InstallPlan
Metadata:
  Creation Timestamp:  2020-01-21T20:13:46Z
  Generate Name:       install-
  Generation:          1
  Owner References:
    API Version:           operators.coreos.com/v1alpha1
    Block Owner Deletion:  false
    Controller:            false
    Kind:                  Subscription
    Name:                  ibm-cp-applications-kabanero
    UID:                   3a5eabd4-ded9-4071-9c69-60c34003ec87
  Resource Version:        266732
  Self Link:               /apis/operators.coreos.com/v1alpha1/namespaces/kabanero/installplans/install-htgqh
  UID:                     1bcef018-a7fa-47a8-8507-57e2660ad37f
Spec:
  Approval:  Automatic
  Approved:  true
  Cluster Service Version Names:
    kabanero-operator.v0.3.4
Status:
  Catalog Sources:
    ibm-cp-applications-catalog
  Conditions:
    Last Transition Time:  2020-01-21T20:13:47Z
    Last Update Time:      2020-01-21T20:13:47Z
[...]

PR #429, opened 2020-01-28
Release v0.3.5-rc.1 which supports OCP 4.3

kaczyns:
The thing that prevents 0.3.4 from working on OCP 4.3, is that the schema in the CRDs is in the wrong spot.  Kube 1.16 validates that if you have only one version in the CRD, that its schema is at the top level.


kaczyns:
#404

PR #442, opened 2020-02-03
stack webhook

dacleyra:


pkg/webhook/stack/validatingwebhook.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	
+	if len(stack.Spec.Versions) == 0 {
+		reason = fmt.Sprintf("Stack Spec.Versions[] list is empty. stack: %v", stack)
+		err = fmt.Errorf(reason)
+		return false, reason, err
+	}
+	
+	for _, version := range stack.Spec.Versions {
+		if (len(version.DesiredState) != 0) && !((version.DesiredState == "active") || (version.DesiredState == "inactive")) {
```
kaczyns: I'm wondering if we should do a case insensitive compare here.  or have the mutating hook fold to lower case and then the validating hook can do a lowercase compare like you're doing here.

pkg/webhook/stack/validatingwebhook.go
```diff
@@ -0,0 +1,106 @@
...
+			reason = fmt.Sprintf("Stack Spec.Versions[].DesiredState may only be set to active or inactive. stack: %v", stack)
+			err = fmt.Errorf(reason)
+			return false, reason, err
+		}
+		
+		if len(version.Images) == 0 {
+			reason = fmt.Sprintf("Stack Spec.Versions[].Images[] list is empty. stack: %v", stack)
```
kaczyns: I think it would be better if this said `Version %v of stack %v must contain at least one entry for spec.versions[x].images` or something like that.  so that it's easier to identify which version of the stack is missing the images.  It should be obvious, but in cases where someone has mis-spelled something, it helps to know which version is wrong.

pkg/webhook/stack/validatingwebhook.go
```diff
@@ -0,0 +1,106 @@
...
+			err = fmt.Errorf(reason)
+			return false, reason, err
+		}
+	
+		for _, pipeline := range version.Pipelines {
+			if len(pipeline.Https.Url) == 0 {
+				reason = fmt.Sprintf("Stack Spec.Versions[].Pipelines[].Https.Url is not set. stack: %v", stack)
```
kaczyns: Again, in these messages, I think we need to be explicit about which version of the stack is wrong.

config/orchestrations/admission-webhook/0.2/kabanero-operator-admission-webhook-config.yaml
```diff
@@ -117,3 +117,63 @@ webhooks:
     scope: '*'
   sideEffects: Unknown
   timeoutSeconds: 30
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: webhook.operator.kabanero.io
+webhooks:
+- admissionReviewVersions:
```
kaczyns: I expected to see this merged in with the `ValidatingWebhookConfiguration` above.  This is a second one, that happens to have the same name as the previous one.

config/orchestrations/admission-webhook/0.2/kabanero-operator-admission-webhook-config.yaml
```diff
@@ -117,3 +117,63 @@ webhooks:
...
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - stacks
+    scope: '*'
+  sideEffects: Unknown
+  timeoutSeconds: 30
+- admissionReviewVersions:
```
kaczyns: This one is not necessary - well I guess it depends on how you merge this validating webhook config with the previous one.

PR #443, opened 2020-02-03
Initial RH-SSO deployment

kaczyns:
This deploys the RH-SSO if the necessary configuration in the Kabanero CR instance is present:
```yaml
spec:
  sso:
    enable: true
    adminSecretName: some-secret
```
There is also a `provider` field which is ignored now, as I suspect that in the future we'll be switching over to Keycloak as an option.  

The secret name has to have a username, password and realm defined in it.

In a future PR, we'll support generating a username/password for the regular user, as well as any other enhancements suggested by the security team.


brutif:
@meiaus fyi

kaczyns:
> @meiaus fyi

Thanks Bruce - I tried to add her as a reviewer but she's not in the kabanero-io org.

kaczyns:
@brutif They have to put adminSecretName and enable.  Provider is not required (and is not checked by anything right now).  I expect that down the road when Kabanero moves back to OKD, we'll not be able to use RHSSO, and we'll have to allow Keycloak as a separate provider.

meiaus:
Thanks Tim.  Bruce as reviewer is sufficient from SSO team.  I've also reviewed internally.

Issue #412, opened 2020-01-23
Configure Codeready Workspaces operator to replace Che


kaczyns:
Please see the details in https://github.com/kabanero-io/roadmap/issues/127.  In addition to what is mentioned there, the Kabanero CR should replace references of che with codeready.

PR #444, opened 2020-02-03
CodeReady Workspaces deployment.

mezarin:
Fixes: #412 (Worked with @johnmcollier  on the details listed below):

Under this PR :
1. Che operator deployment is removed.
2. The codeready-workspaces (CRW) operator and CRW CR instance are created in the kabanero namespace.
3. The CRW Operator is installed by the installation script by default. The operator version installed is the version associated with the OLM subscription channel named 'latest'. It is version 2.0.0 currently.   
4. The CRW instance is created by the Kabanero CR instance, but CRW instance creation is disabled by default. The updated Kabanero model shown below shows how to enable it. 
5. When the CRW instance is deployed, the following defaults are used if they are not especified in the kabanero instance yaml:

```
devfileRegistryImage: 'kabanero/che-devfile-registry:0.8.0'
tlsSupport: false
selfSignedCert: false
cheWorkspaceClusterRole: 'eclipse-codewind'
openShiftoAuth: false
```
These values are configurable through the Kabanero instance deployment yaml. 

For example:

```
apiVersion: kabanero.io/v1alpha2
kind: Kabanero
metadata:
  name: kabanero
spec:
  version: "0.6.0"
  codeReadyWorkspaces:
    enable: true
    operator:
      customResourceInstance:
        devFileRegistryImage:
          version: "0.6.0"
          repository: kabanero/che-devfile-registry
          tag: "0.8.0"
          image: kabanero/che-devfile-registry:0.8.0
        cheWorkspaceClusterRole: eclipse-codewind
        tlsSupport: false
        selfSignedCert: false
        openShiftOAuth: false 

```
The customization of devFileRegistryImage follow the kabanero operator image versioning precedence rules. That is specifying the image takes precedence over specifying repository + tag.
 
Notes:
- The user is responsible to make any required cluster/config/etc updates needed prior to setting the value of selfSignedCert, openShiftoAuth, and/or tlsSupport to true. 
- Eclipse-che operator and Codeready-workspaces operator cannot be both installed at the same time in the same namespace. The kabanero operator installation will attempt to uninstall the eclipe-che CR instance (codewind-che) and Eclipse-che operator previously installed by the kabanero operator. 
If the current installation of Eclipse-che was not done through the kabanero operator/install, it must be uninstalled prior to running the kabanero installation script. Furthermore, users need to be sure to delete any operatorgroups other than kabanero (i.e. kabanero-5pn6b) in the kabanero namespace that may have been created as result of a manual Eclipse-che installation.


config/orchestrations/codeready-workspaces/0.1/codeready-workspaces-cr.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: org.eclipse.che/v1
+kind: CheCluster
+metadata:
+  # The name prefixed with the kabanero instance name for multiple instance creation.
+  name: {{ .kabaneroInstanceName }}-codeready-workspaces
```
kaczyns: We can't do this, because there is a name length limit in Kubernetes.  Having said that, there can only be a single `Kabanero` CR instance in a namespace, so I don't think we really need to pre-pend the name with the Kabanero instance name.

config/orchestrations/codeready-workspaces/0.1/codewind-clusterrole.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
kaczyns: I don't understand why this is a cluster role.  Che/CRW is installed to a single namespace, right?  Can this just be a Role that is bound to this one namespace?

What happens if I have a second Kabanero instance in a second namespace?  How are we preventing the ClusterRole from being deleted if the second Kabanero instance is deleted?  Wouldn't the first Che/CRW instance stop working?

config/versions.yaml
```diff
@@ -17,12 +17,12 @@ kabanero:
   related-versions: 
     cli-services: "0.5.0"
     landing: "0.5.0"
-    kabanero-che: "0.6.0"
```
kaczyns: Can we delete the orchestrations for kabanero-che now?

config/versions.yaml
```diff
@@ -46,6 +46,13 @@ related-software:
       repository: "kabanero/kabanero-che"
       tag: "0.6.0"
 
+  codeready-workspaces:
+  - version: "0.6.0"
```
kaczyns: I would prefer to see this version number match the version number of the operator.  The eclipse-che version was 0.6.0 long before Kabanero 0.6.0 existed, it just happened to catch up to the Kabanero number.  I don't think we intend to have a version of the devfile per kabanero release, right?  Maybe using the devfile release number would be more appropriate.

deploy/install.sh
```diff
@@ -207,14 +257,24 @@ oc apply -f $KABANERO_SUBSCRIPTIONS_YAML --selector kabanero.io/install=13-subsc
 checksub openshift-pipelines openshift-operators
 checksub appsody-operator-certified openshift-operators
 
-# Install 14-subscription (che, kabanero)
+# Install 14-subscription (codeready-workspaces, kabanero)
+
+# Prior to installing codeready-workspaces, remove all eclipse-che instances and 
+# eclipse-che operator deployments in the kabanero namespace.
+removeResourceInstance checlusters.org.eclipse.che CheCluster kabanero
+unsubscribe eclipse-che kabanero
```
kaczyns: I am just wondering, if we're going to have to go thru and clean up all of the stale CSVs..... hmm no we're not because Che was just installed at the namespace scope.  I guess we should just keep this in mind for the future, if we ever remove a cluster scope operator in the install script, there is a boat-load of extra work to be done to clean up the mess.

deploy/install.sh
```diff
@@ -207,14 +257,24 @@ oc apply -f $KABANERO_SUBSCRIPTIONS_YAML --selector kabanero.io/install=13-subsc
...
+
+# Codewind is required to run as privileged and as root because it builds container images
+oc adm policy add-scc-to-user anyuid system:serviceaccount:kabanero:che-workspace
+oc adm policy add-scc-to-user privileged system:serviceaccount:kabanero:che-workspace
```
kaczyns: Is there something fundamentally different about using CRW instead of Che, with regards to these two `oc adm` commands?  Why didn't we have to do this when we were using the Che operator?

pkg/apis/kabanero/v1alpha2/kabanero_types.go
```diff
@@ -260,30 +268,26 @@ type KappnavStatus struct {
...
-	KabaneroCheInstance KabaneroCheInstanceStatus `json:"kabaneroCheInstance,omitempty"`
-}
-
-// CheOperatorStatus defines the observed status details of the Che operator.
-type CheOperatorStatus struct {
-	Version string `json:"version,omitempty"`
+// CRWStatus defines the observed status details of codeready-workspaces.
+type CRWStatus struct {
+	Ready        string            `json:"ready,omitempty"`
+	ErrorMessage string            `json:"errorMessage,omitempty"`
```
kaczyns: This should probably change to just `Message` right?

deploy/install.sh
mezarin: Per the codewind documentation (https://www.eclipse.org/codewind/mdt-che-installinfo.html), setting the adm policies are/were required from a codewind point of view. It was just not done as part of the che installation, and left to the user to do that. Since this is explicitly required now, perhaps it should have been done for che.

config/orchestrations/codeready-workspaces/0.1/codewind-clusterrole.yaml
kaczyns: Ed explained that there is some logic in the Kabanero finalizer that will prevent the cluster role from being deleted if there is another Kabanero CR instance on the system.

John explained that Che expects this to be a cluster role, but we could manually create a role and a role binding (and ignore the cluster role setting on the CheCluster).

For the time being we are just going to leave it as it is.  When we start officially supporting two Kabanero instances on the system, if the Kabanero operator hasn't become a cluster scoped operator yet, we'll consider changing this to a Role.

config/versions.yaml
kaczyns: Ed and I couldn't come up with a more appropriate identifier, so we're just going to leave it like this for now.

PR #449, opened 2020-02-04
stack version semver check

dacleyra:


pkg/webhook/stack/validatingwebhook.go
```diff
@@ -71,6 +73,13 @@ func (v *stackValidator) validateStackFn(ctx context.Context, stack *kabanerov1a
 			err = fmt.Errorf(reason)
 			return false, reason, err
 		}
+
+		_, err := semver.Parse(version.Version)
+		if err != nil {
+			reason = fmt.Sprintf("Stack %v %v spec.Versions[].Version must be semver. %v. stack: %v", stack.Spec.Name, version.Version, err, stack)
```
kaczyns: I think the semver parser that we are using is pretty strict.  I think at the very least we need to say in the message that we require the version to have three parts (major.minor.patch).  I know that semver requires this, but it's one those things where some people think they are using semver even though it just has major.minor.  (another one is people who put the `v` on the front, like `v0.0.1`).  

Can we change the message to say "Stack %v %v spec.Versions[].Version must conform to semver major.minor.patch.  %v. stack: %v"  ?

pkg/webhook/stack/validatingwebhook_test.go
```diff
@@ -225,6 +225,28 @@ func TestValidatingWebhook9(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("Validation failed. An error was expected: ", err)
+	}
+}
+
+
+// Spec.Versions not semver
+func TestValidatingWebhook10(t *testing.T) {
+	newStack := validatingStack.DeepCopy()
+	newStack.Spec.Versions[0].Version = "1.0"
```
kaczyns: Your test case illustrates perfectly how we can educate someone who really only understands semver a little bit, what they need to do to be compliant here.  1.0 should be 1.0.0.

pkg/webhook/stack/validatingwebhook.go
kaczyns: Dan looked up the message we get back from the semver parser and it's already doing a good job of telling the user what is wrong.

PR #460, opened 2020-02-07
go 1.13, operator-sdk 0.15.1

dacleyra:
Update build to go 1.13 and operator-sdk 0.15.1
migrate from 	operator-sdk generate openapi to operator-sdk generate crds
Trim outdated deps make target
fix typo in admission-webhook url

Update go.mod replaces to be in-line with operator-sdk 0.15.1
go-autorest up to v13.3.3 or else build failure

Set // +kubebuilder:storageversion on kabanero/v1alpha2 (required to build)

// StorageVersion marks this version as the "storage version" for the CRD for conversion.
//
// When conversion is enabled for a CRD (i.e. it's not a trivial-versions/single-version CRD),
// one version is set as the "storage version" to be stored in etcd.  Attempting to store any
// other version will result in conversion to the storage version via a conversion webhook.


kaczyns:
@dacleyra I added a commit and changed the way we're getting around the go-autorest problem.  When I do `go mod tidy` it removes the `require go-autorest` from `go.mod` because our project doesn't actually use go-autorest.  After way too much time searching, the internet told me to use `go mod graph` to see who is importing the conflicting versions.  In the end requiring a specific version of promethius leaves everyone using the same version of go-autorest.

PR #462, opened 2020-02-07
fix db volume mount

brutif:
fixes a bug where the db is writing to /var/lib/pgsql/data but pv is mounted to /var/lib/pgsql/data:z which prevents data from being persisted across pod restarts.

### DIFF
--- a/config/orchestrations/sso/0.1/sso.yaml
+++ b/config/orchestrations/sso/0.1/sso.yaml
@@ -273,7 +273,7 @@ spec:
           initialDelaySeconds: 5
           timeoutSeconds: 1
         volumeMounts:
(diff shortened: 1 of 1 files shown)


config/orchestrations/sso/0.1/sso.yaml
```diff
@@ -287,7 +287,7 @@ spec:
       - sso-postgresql
       from:
         kind: ImageStreamTag
-        name: postgresql:10
+        name: postgresql:9.6
```
kaczyns: I'm asking @brutif if this change was intentional as it's not mentioned in the PR.

brutif: yes

PR #481, opened 2020-02-12
Use Github APIs to download stack index file.

### LINKED ISSUE #469: Retrieve release assets from GHE
Given a Github HTTP client (see #467 ) retrieve a release asset from GHE.  The asset could be a yaml file or a zipped tarball (index or pipeline zip).

The following modifications should be made to the Kabanero CRD to accommodate for the additional parameters necessary when using the Github API (see this example CR instance):
```yaml
apiVersion: kabanero.io/v1alpha2
kind: Kabanero
metadata:
  name: kabanero
spec:
  version: "0.6.0"
  stacks: 
    repositories: 
      - name: central
        githubRelease:
          hostname: github.mydomain.com
          organization: myOrg
          project: myProject
          release: myRelease
          assetName: incubator-index.yaml
```
Note that the `githubRelease` is mutually exclusive with the existing `https`.  Only one or the other should be specified.  The `githubRelease` stanza includes all of the information necessary to use the github API to retrieve a release asset.  We could consider having another way of specifying all of this information (ie a `url` field that specifides `https://github.mydomain.com/myOrg/myProject/releases/download/myRelease/incubator-index.yaml`) but for now lets list it out explicitly as the github API requires it.

This is the Github Go client:
https://godoc.org/github.com/google/go-github/github#Client

The `Repositories` member provides access to the releases and their assets.  The following appears to download a release asset:
https://godoc.org/github.com/google/go-github/github#RepositoriesService.DownloadReleaseAsset

You need to know the `id` of the asset that you want to get.  This may require some digging:
https://godoc.org/github.com/google/go-github/github#RepositoriesService.GetReleaseByTag
to get the release ID. 
https://godoc.org/github.com/google/go-github/github#RepositoriesService.ListReleaseAssets
to match the asset name to the asset ID.

If there is an easier way to do this, I'm all for it.  I don't think we can just jam a request in for the whole URL, but if we can we should do that, it's probably easier.  That is, just use the Github HTTP client but pass in the URL to the asset that we want.  I don't think that will work, but if it does, great.

Another issue will integrate the HTTP request/response that we use to retrieve the release asset, with the cache that we maintain.  This assumes that Github continues to use `etags` when using this API.

mezarin:
Fixes #469


pkg/controller/stack/resolver.go
```diff
@@ -108,3 +117,109 @@ func SearchStack(stackName string, index *Index) ([]Stack, error) {
...
+	// Get a Github client.
+	gclient, err := getGitClient(repoConf.GitRelease)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the release tagged in Github as repoConf.GitRelease.Release.
+	release, response, err := gclient.Repositories.GetReleaseByTag(context.Background(), repoConf.GitRelease.Organization, repoConf.GitRelease.Project, repoConf.GitRelease.Release)
+	if err != nil || response.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Unable to retrieve object representing Github repository release %v. Configured GitRelease data: %v. Error: %v", repoConf.GitRelease.Release, repoConf.GitRelease, err)
```
kaczyns: Should we report the HTTP status code here too?

pkg/controller/stack/resolver.go
```diff
@@ -108,3 +117,109 @@ func SearchStack(stackName string, index *Index) ([]Stack, error) {
...
+	}
+	assets := release.Assets
+
+	// Find the asset identified as repoConf.GitRelease.AssetName and download it.
+	for _, asset := range assets {
+		if asset.GetName() == repoConf.GitRelease.AssetName {
+			id := asset.GetID()
+			reader, redirectURL, err := gclient.Repositories.DownloadReleaseAsset(context.Background(), repoConf.GitRelease.Organization, repoConf.GitRelease.Project, id)
```
kaczyns: this is fine... but I thought the github client would let you follow re-directs if you supplied an HTTP client to do so with.  Did you try this?
https://github.com/google/go-github/commit/75ddd505ad1c4d7c2d0bd0c5e0ef4e753907fe6a

mezarin: The code is passing an HTTP client. It is the default one, but a client nonetheless. The case that is documented is the one where you get a redirect if the value of the client is nil.

pkg/controller/stack/resolver.go
mezarin: Yeah it would be helpful if it had that in the message.

pkg/controller/stack/resolver.go
kaczyns: I don't see where it's passing a redirect client.  the args are context, org, project, and id.  there should be a 5th arg, which is the client to use for redirects.  unless the code defaults to use the default client if one is not provided (I didn't think that it did).  Is that what is happening here?

mezarin: Ok, i got to the bottom of this. It turns out that we were looking at two different versions of the doc/code. The level of go-github imported by VS code is not the very latest one. The change in go-github that introduced the redirect logic you are talking about made it into a release 3 days ago. The doc you are looking at was, most likely, out there around 25 days ago. So that explains why i HAD to handle the redirects myself and why you thought i didn't need to. I will try to override the go-github version with the very latest and see how it goes.

Issue #483, opened 2020-02-12
0.6.0 pipelines are not being recreated when deleted


dacleyra:
Kabanero configured with full.yaml
Delete a pipeline
Expect the pipeline to be recreated

kaczyns:
We'll need to check the watches that the stack controller is setting up.  I would not be surprised if it was not watching for pipelines and tasks.  And it can't watch for trigger templates or bindings in the tekton-pipelines namespace (at least it couldn't when we were using controller-runtime 0.1.x... maybe it can now)

marikaj123:
@dacleyra - Can you please provide us with a status on this issue?

dacleyra:
A PR is ready for review

marikaj123:
@dacleyra -Is the defect ready to close?

dacleyra:
PR was merged, closing

Issue #491, opened 2020-02-14
stack-controller should watch TriggerBinding and TriggerTemplate


dacleyra:
Watch and reconcile changes/deletions to owned TriggerBinding and TriggerTemplate, as is done for Pipeline & Task to ensure integrity.

Granted, #483 needs to be fixed

kaczyns:
We've already discussed this internally, but I'll just write it down for other's awareness.  We can't fix this issue until we have a cluster scoped operator.  This will probably happen when we implement #284 .

kaczyns:
I am going to lump this into #284 for now.  However I'd like to start by keeping the stack controller namespace-scoped, and just resolving the issue as it's described above, by watching objects in the other namespace.  We can continue to have a discussion about whether the stack controller should be namespace scoped or cluster scoped.  I am not unwilling to change my opinion on what the scoping should be.  My thinking on this right now is as follows:
1) @dacleyra discovered that it is possible to have the manager cache for multiple namespaces (see https://github.com/operator-framework/operator-sdk/blob/v0.17.x/doc/user-guide.md#manager).  The Kabanero/Team controller should be able to set an env var in the stack controller that tells it what to watch.  In theory this should be enough to resolve this issue.
2) Having a cluster scoped Stack controller would make it impossible to start a version of the stack controller that differs from the version of the Kabanero operator, or at least would make it impossible to have two different Kabanero/Team CR instances at different versions of Kabanero.  There are downstream requirements that the operator be capable of installing a certain number of previous versions of Kabanero, to allow users to upgrade at their own pace.  
3) The stack controller interacts with some of the other versioned components.  For example as-is you can say `spec.version: "0.9.0"` in your Kabanero CR instance, and you get the "0.9.0" versions of the CLI service, landing page, events operator, and stack controller.  The stack controller works with these other components.  If we forced the stack controller to be the version of the kabanero operator, it might not be compatible with version "0.9.0" of the events operator.  This is a somewhat arbitrary line since one could say the kabanero/team controller might not be compatible with the stack controller, but that same incompatibility would apply to every versioned component and in the event it was not compatible, the kabanero/team controller would just not allow you to install that version.

So to fix this issue, the kabanero/team controller needs to set an environment variable in the stack controller deployment/pod, telling it what additional namespaces need to be watched.  For the time being it's the `tekton-pipelines` namespace, since that's where the trigger bindings and trigger templates are.  In the future it will also be the `pipelinesNamespace` but we're not there yet (#230).

PR #492, opened 2020-02-14
Watch Conditions, TriggerBindings, TriggerTemples

dacleyra:
This should fix issue #483 and work towards fixing issue #491 

For created assets (tasks/pipelines/triggers) set ownerReference controller=true.
In turn they are properly watched and recreated if modified/deleted.

Add Watch for Condititions

Add Watch for TriggerBindings, TriggerTemplates...

It turns out the TriggerBindings and TriggerTemplates can't really be watched  unless Kabanero is a cluster operator. The objects are created in `tekton-pipelines` namespace, but the operator env WATCH_NAMESPACE is scoped only to `kabanero`. They don't get an ownerRef.

Should probably be picked for 0.6

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -308,8 +308,8 @@ oc new-project tekton-pipelines || true
 
 openshift_master_default_subdomain=$(oc get ingresses.config.openshift.io cluster --output=jsonpath={.spec.domain})
 
-curl -s -L https://github.com/tektoncd/dashboard/releases/download/v0.5.1/openshift-tekton-webhooks-extension-release.yaml | sed "s/{openshift_master_default_subdomain}/${openshift_master_default_subdomain}/" | oc apply -f -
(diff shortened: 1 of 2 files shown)


registry/manifests/kabanero-operator/0.6.0/kabanero-operator.v0.6.0.clusterserviceversion.yaml
```diff
@@ -178,6 +178,15 @@ spec:
         - kind: Task
           name: ''
           version: v1alpha1
+        - kind: Condition
+          name: ''
+          version: v1alpha1
+        - kind: TriggerBinding
```
kaczyns: The trigger bindings and trigger templates don't follow the object ownership model (they are in a different namespace) so there will never be any of these objects in the openshift console.  It's probably best to leave them off of the resources view (unless there's another reason why they are here, in which case please let me know what that is).

PR #496, opened 2020-02-17
liveness probe to /health

dacleyra:
This should be picked for 0.6 as well


mezarin:
Dan(@dacleyra), CLI services is fixing this issue under: https://github.com/kabanero-io/kabanero-command-line-services/pull/120. They are going to re-cut their *rc.4 release soon. We should hold off merging this PR.

dacleyra:
Ok, looks like they updated their image. This was the wrong solution.

PR #500, opened 2020-02-18
servicemeshoperator-1.0.7

dacleyra:


jgawor:
@kaczyns @dacleyra servicemesh is GAed. why maintain it in the kabanero catalog?

PR #508, opened 2020-02-20
Use manifestival/manifestival instead of kabanero-io/manifestival

kaczyns:
The original Manifestival library now supports creating manifests from other source types.  Since that was essentially the only difference between our fork and the original, we should stop using our fork.

These were the changes that pertained to Kabanero:
* Manifest objects are now immutable, so we receive a new object after Transform()
* We're using the Manifestival controller-runtime client
* ApplyAll and DeleteAll were renamed to Apply and Delete respectively

@jcrossley3 if you have a few minutes to look this over, we'd appreciate it!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,9 +8,10 @@ require (
 	github.com/go-openapi/spec v0.19.4
 	github.com/google/go-cmp v0.3.1
 	github.com/google/go-github/v29 v29.0.3
-	github.com/kabanero-io/manifestival v0.0.0-20191230200607-194733cd27d9
 	github.com/knative/pkg v0.0.0-20190817231834-12ee58e32cc8 // indirect
 	github.com/knative/serving-operator v0.0.0-20190702004031-e30377b852ff
+	github.com/manifestival/controller-runtime-client v0.1.1-0.20200218204725-1af9550ddf8f
+	github.com/manifestival/manifestival v0.1.1-0.20200219193505-fabb889b98f5
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
 	github.com/operator-framework/operator-lifecycle-manager v3.11.0+incompatible
 	github.com/operator-framework/operator-sdk v0.15.1
--- a/pkg/controller/kabaneroplatform/admission_controller_webhook.go
+++ b/pkg/controller/kabaneroplatform/admission_controller_webhook.go
@@ -85,7 +85,7 @@ func reconcileAdmissionControllerWebhook(ctx context.Context, k *kabanerov1alpha
 		return err
 	}
 
-	mOrig, err := mf.ManifestFrom(mf.Reader(strings.NewReader(s)), mf.UseClient(mfc.NewClient(c)))
+	mOrig, err := mf.ManifestFrom(mf.Reader(strings.NewReader(s)), mf.UseClient(mfc.NewClient(c)), mf.UseLogger(reqLogger.WithName("manifestival")))
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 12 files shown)


pkg/controller/kabaneroplatform/landing.go
```diff
@@ -203,18 +204,18 @@ func cleanupLandingPage(k *kabanerov1alpha2.Kabanero, c client.Client) error {
...
-	err = m.Transform(transforms...)
+	m, err := mOrig.Transform(transforms...)
 	if err != nil {
 		return err
 	}
 
-	err = m.DeleteAll()
+	err = m.Delete()
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
```
jcrossley3: This is redundant. The default behavior of `m.Delete()` will "eat" the `NotFound` error. There is an option to override that behavior, but I think most folks just want the manifest resources gone.

go.mod
```diff
@@ -8,9 +8,10 @@ require (
 	github.com/go-openapi/spec v0.19.4
 	github.com/google/go-cmp v0.3.1
 	github.com/google/go-github/v29 v29.0.3
-	github.com/kabanero-io/manifestival v0.0.0-20191230200607-194733cd27d9
 	github.com/knative/pkg v0.0.0-20190817231834-12ee58e32cc8 // indirect
 	github.com/knative/serving-operator v0.0.0-20190702004031-e30377b852ff
+	github.com/manifestival/controller-runtime-client v0.1.1-0.20200218204725-1af9550ddf8f
+	github.com/manifestival/manifestival v0.1.1-0.20200219193505-fabb889b98f5
```
jcrossley3: I guess I need to bite the bullet and figure out go modules. :smile: I'm confused by the release version since I don't recall releasing that version. I'm currently struggling with versions trying to port one of our operator-sdk-generated apps, so I'm guessing you're still using a pretty old (or much newer?) version.

pkg/controller/kabaneroplatform/admission_controller_webhook.go
```diff
@@ -84,7 +85,7 @@ func reconcileAdmissionControllerWebhook(ctx context.Context, k *kabanerov1alpha
 		return err
 	}
 
-	m, err := mf.FromReader(strings.NewReader(s), c)
+	mOrig, err := mf.ManifestFrom(mf.Reader(strings.NewReader(s)), mf.UseClient(mfc.NewClient(c)))
```
jcrossley3: One other note: you may want to pass `mf.UseLogger(reqLogger.WithName("manifestival"))` or similar to have log what it's doing. Nice choice using `go-logr`, btw!

pkg/controller/kabaneroplatform/landing.go
kaczyns: Great - I like that default behavior :smile:

go.mod
kaczyns: Well, let me start by saying that we are still pretty clueless about modules.  We were more or less forced to switch because operator-sdk switched (or at least that's how we understood it to be).  We're on operator-sdk 0.15.1 now, which is pretty current.

In this specific case, we tried to let `go` figure out what version to use, and it picked the latest release in `manifestival/manifestival` which was v0.1.0.  However that didn't have the changes you made to allow multiple sources.  So we've learned that you can tell `go.mod` to use the `master` branch, and then `go` will generate some string that somehow identifies what that is.... in this case I'm assuming it picked `v0.1.1` because that would come next in semver, and then the date and commit hash from where master was at that point in time.  Maybe?

pkg/controller/kabaneroplatform/admission_controller_webhook.go
kaczyns: Excellent - we'll do that too.

go.mod
jcrossley3: Ah, good to know! I was planning to cut a 0.2.0 after I get my dep issue solved, but I guess as long as you have a recent master revision (whatever go.mod wants to call it) working for you, it doesn't matter that much.

BTW, I'm trying to be better about communicating which features are included in which release [here](https://github.com/manifestival/manifestival/blob/master/CHANGELOG.md).

kaczyns:
@mezarin @dacleyra If one (or both) of you would review and merge this, I'd appreciate it.  

Well I guess only one of you can merge it.  Technically.

PR #530, opened 2020-02-27
csv upg fix

dacleyra:
Include the missing 0.3.5 csv into registry
correct the replaces for 0.4.0 so the olm graph can build the upgrade from 0.3.5
fix the image tag on previous release csvs


registry/manifests/kabanero-operator/0.4.0/kabanero-operator.v0.4.0.clusterserviceversion.yaml
```diff
@@ -52,7 +52,7 @@ spec:
   apiservicedefinitions: {}
   maturity: alpha
   version: 0.4.0
-  replaces: kabanero-operator.v0.3.1
+  replaces: kabanero-operator.v0.3.5
```
kaczyns: This is not going to hurt anything, but it's also not necessary, because we also specify a skiprange on the next minor version.

Issue #545, opened 2020-03-03
Invalid image value


jgawor:
my kabanero-index has the following:
```
apiVersion: v2
stacks:
- id: java-microprofile
  name: Eclipse MicroProfile®
  version: 0.2.25
  description: Eclipse MicroProfile on Open Liberty & OpenJ9 using Maven
  license: Apache-2.0
  language: java
  maintainers:
  - name: Emily Jiang
    email: emijiang6@googlemail.com
    github-id: Emily-Jiang
  - name: Neeraj Laad
    email: neeraj.laad@gmail.com
    github-id: neeraj-laad
  - name: Ozzy
    email: ozzy@ca.ibm.com
    github-id: BarDweller
  default-template: default
  src: http://kabanero-index-kabanero.apps.lisa7.gawor.me/java-microprofile.v0.2.25.source.tar.gz
  templates:
  - id: default
    url: http://kabanero-index-kabanero.apps.lisa7.gawor.me/java-microprofile.v0.2.25.templates.default.tar.gz
  requirements:
    appsody-version: '>= 0.5.0'
  image: image-registry.openshift-image-registry.svc:5000/kabanero/java-microprofile:0.2.25
```
The CR is configured to point at this index. The operator created the following stack for it:

```
$ oc get stack java-microprofile -o yaml
apiVersion: kabanero.io/v1alpha2
kind: Stack
metadata:
  creationTimestamp: "2020-03-03T02:41:46Z"
  finalizers:
  - kabanero.io/stack-controller
  generation: 1
  name: java-microprofile
  namespace: kabanero
  ownerReferences:
  - apiVersion: kabanero.io/v1alpha2
    controller: true
    kind: Kabanero
    name: kabanero
    uid: 2cd39aa8-b4d1-4537-99f1-c17405904d9f
  resourceVersion: "1214352"
  selfLink: /apis/kabanero.io/v1alpha2/namespaces/kabanero/stacks/java-microprofile
  uid: 0530a0ab-6d02-477f-8ee2-55ca88828271
spec:
  name: java-microprofile
  versions:
  - images:
    - id: Eclipse MicroProfile®
      image: image-registry.openshift-image-registry.svc
    pipelines:
    - https:
        url: https://github.com/kabanero-io/kabanero-pipelines/releases/download/0.6.0/default-kabanero-pipelines.tar.gz
      id: default
[...]

PR #554, opened 2020-03-04
v0.6.1-rc.1

kaczyns:
Pull in fix for #545


kaczyns:
I added a 2nd commit, that pulls in service mesh 1.0.8

PR #557, opened 2020-03-05
Ignore colons in the first token of an image repository

kaczyns:
There are multiple places in Kabanero where we attempt to remove a tag from an image name.  We can argue whether there should be a single place where we do this, but, the operation should be idempotent.

If a colon occurs in the first path component of a multi-component container image name, it is not the tag separator and the Kabanero operator should ignore it.

@jgawor I would appreciate your review as well.


jgawor:
LGTM

Issue #353, opened 2019-12-10
Correct the behavior of desiredState in the Stack CR


kaczyns:
The operator's implementation of `desiredState` in the Stack CR instance drifted from its intended behavior.  The `desiredState` was meant to be an over-ride used by the Kabanero CLI.  The current implementation uses it as a general purpose activate/deactivate switch.  The following changes should be considered:
1) The `desiredState` field of the Stack CR should be an over-ride and should be the empty string by default.
2) The stack controller should treat an empty `desiredState` as `activate`, and should activate the stack.
3) The Kabanero controller should always set the `desiredState` to the empty string when it creates a stack, or when it adds a version of a stack to the versions array of the Stack CR instance.
4) The Kabanero controller should ensure that the set of Stack CR instances matches exactly the view of the Stack Hubs configured in the Kabanero CR instance, unless a Stack CR instance has the `desiredState` field set to a non-empty string.  In this case, the Kabanero controller should not alter that stack version or Stack CR.  Note that within a Stack CR, some versions may have the `desiredState` field set, and some may not.

This will result in a behavior change, and I would suggest having a `v1alpha2` version of the Kabanero CRD which eliminates the `activateDefaultCollections` field.

kaczyns:
I am working on this over here: https://github.com/kaczyns/kabanero-operator/tree/v1alpha2-sdk11
This is going to need to be staged - I wanted to remove the requirement that the Collection CR instance has a Kabanero owner at the same time, but I don't think that will be possible.  I'd like to wait until the current backlog of PRs is reviewed / merged before making more progress here.

marikaj123:
The release was originally set to 0.6.0 and I have moved it to a later one.   Please feel free to to reset it, as needed.

kaczyns:
We need to remove the note that was added in kabanero-io/docs#338 now that we've changed the behavior of desiredStatre.  The kabanero operator is now deleting stacks that are no longer present in the index (unless the desiredState is set to a non-null value).

PR #562, opened 2020-03-06
Cleanup existing stacks

mezarin:
Associated with #353.
This PR allows existing stacks to be cleaned up based on their set desired state (existing stack versions with an active desired state are not touched per the new definition of desired state - see 353 for details) and the index file(s) being processed.

### DIFF
--- a/pkg/controller/kabaneroplatform/featured_stacks.go
+++ b/pkg/controller/kabaneroplatform/featured_stacks.go
@@ -81,8 +81,9 @@ func reconcileFeaturedStacks(ctx context.Context, k *kabanerov1alpha2.Kabanero, 
 			for j, stackVersion := range stackResource.Spec.Versions {
 				if stackVersion.Version == stack.Version {
 					foundVersion = true
-					// Per the new defintion of active, do not update any existing stacks with matching names and versions.
-					if !(alreadyDeployed && stackVersion.DesiredState == "active") {
+					// Per the new defintion of desired state, do not update any existing stacks if the desired state is set
+					// to an allowed value.
+					if !(alreadyDeployed && len(stackVersion.DesiredState) > 0) {
 						stackVersion.Pipelines = stack.Pipelines
 						stackVersion.SkipCertVerification = stack.SkipCertVerification
 						stackVersion.Images = stack.Images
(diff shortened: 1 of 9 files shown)


pkg/controller/kabaneroplatform/featured_stacks.go
```diff
@@ -72,10 +81,13 @@ func reconcileFeaturedStacks(ctx context.Context, k *kabanerov1alpha2.Kabanero,
 			for j, stackVersion := range stackResource.Spec.Versions {
 				if stackVersion.Version == stack.Version {
 					foundVersion = true
-					stackVersion.Pipelines = stack.Pipelines
-					stackVersion.SkipCertVerification = stack.SkipCertVerification
-					stackVersion.Images = stack.Images
-					stackResource.Spec.Versions[j] = stackVersion
+					// Per the new defintion of active, do not update any existing stacks with matching names and versions.
+					if !(alreadyDeployed && stackVersion.DesiredState == "active") {
```
kaczyns: I'm wondering if this should be `len(stackVersion.DesiredState) > 0` instead of `stackVersion.DesiredState == "active"`.  If someone sets this to "inactive" I think we'll be changing the pipelines and images in the Stack CR.  The user may not be expecting that, since they set `desiredState` to a non-null value.

pkg/controller/kabaneroplatform/featured_stacks.go
```diff
@@ -137,3 +148,53 @@ func featuredStacks(k *kabanerov1alpha2.Kabanero, cl client.Client) (map[string]
...
+			for _, iStack := range iStackList {
+				if dStackVersion.Version == iStack.Version {
+					deployedStackVersionMatchIndex = true
+					newStackVersions = append(newStackVersions, dStackVersion)
+					break
+				}
+			}
+
+			// Keep any stack versions that have a desired state of active.
+			if !deployedStackVersionMatchIndex && dStackVersion.DesiredState == kabanerov1alpha2.StackDesiredStateActive {
```
kaczyns: Ok so now that I see this too, I wonder if I just did a poor job of explaining desiredState.  I think that if desiredState is set to anything (not just `active`) that the operator should not delete it.

pkg/controller/kabaneroplatform/featured_stacks_test.go
```diff
@@ -509,3 +568,330 @@ func TestResolveFeaturedStacksTwoRepositories(t *testing.T) {
...
 }
+
+// Tests that if an existing stack version with an non-active desired state matches the name/version of a stack in the index, the
+// data associated with the index version overrides the existing version.
+// Tests that a stack version with an unset desired state is removed if it is not found in the index being deployed.
+// Tests that if an existing stack version with an active state is not found in the index being deployed, the existing version
+// remains untouched.
+func TestResolveFeaturedStacksCleanupMatchingStackVersionWithNonActiveState(t *testing.T) {
+	stack := stackResource.DeepCopy()
+	stack.Spec.Versions[0].DesiredState = "NotAValidState"
```
kaczyns: This test probably gets simpler if we treat all non-null desiredState values the same.

mezarin:
Hey Tim, Our conversation has heavily focused an active state, which may have made me miss the point that the current allowed desiredState values, from a user perspective, mean what they used to with the difference that now it also means that if any of the allowed values are set, the operator will not update/delete the stack.
Anyway, i made some updates. Let me know if something else needs to change or if we are still not on the same page.
Thanks.

Issue #468, opened 2020-02-10
Select the correct secret for accessing a github server


kaczyns:
Assume that the user has created one or more secrets in the Kabanero control namespace (ie `kabanero` namespace) that contain:
1) an annotation that describes this as a secret that should be used by Kabanero to access github, of the form `kabanero.io/git-0: https://github.mydomain.com`
2) a key in the secret named `password` which represents a Github PAT

Create a utility function that takes a hostname as input, and matches it to a secret.  The following rules can be used to help match the secret to the annotation:
1) the hostname should match exactly, discarding any path that might exist after it (for now)
2) if there are multiple matches (multiple secrets that have a matching annotation) then the annotation with the lexically lowest value key should be used (ie `kabanero.io/git-1` is lower than `kabanero.io/git-9`).
3) if a secret exists with an annotation that matches with no key (ie `kabanero.io/git-0: `) then it should be used if there are no other matching secrets

A future implementation may allow for an added path element to the url, to allow a different secret for say `https://github.mydomain.com/org1` vs `https://github.mydomain.com/org2`.  It would be nice if this implementation did not preclude that future implementation from being easily achievable, but the focus here is on the initial implementation.

A pointer to the matching secret (the whole thing) should be returned so that other code can extract the necessary key/value for use in authenticating to Github.

PR #566, opened 2020-03-12
Report Stack image digest in status

### LINKED ISSUE #437: Retrieve a stack container image digest from its registry and copy to status
A stack index will contain an `image:` in its yaml which describes where that stack's container lives.  The Stack controller should attempt to look up the Sha256 (digest) of that container when the stack is activated, and copy the digest to the `status`.  

Later, during a pipeline run, a validate step in the pipeline will retrieve the digest of the container that is being used in the pipeline run, and compare it to the digests for the active stacks, to see if the pipeline run should be allowed to proceed.

This use of the digest prevents cases where the app developer has coded something like `stack:0.2` into their application, and the cluster has `stack:0.2.8` installed.  If the two share the same digest, we can allow the pipeline run to proceed.  However, if `stack:0.2.9` is available, but not installed, and it's also linked to `stack:0.2`, then the pipeline run will fail unless either the developer updates their app to use `stack:0.2.8`, or `stack:0.2.9` is installed on the cluster.

mezarin:
Fixes #437.

New Kabanero CR instance spec entry:

Example:
(Allowed values: strictDigest, activeDigest, ignoreDigest, and none)
```
spec:
  governancePolicy:
    stackPolicy: activeDigest
```

Sample showing the activation digest in stack's status:

```
status:
  versions:
    - images:
        - digest:
            activation: 9e69ab0182cf6301867f5b597828b35a72d15791aafe54fd8396ff2d7b0b9acb
          id: Eclipse MicroProfile®
          image: docker.io/kabanero/java-microprofile
...
```
```
status:
  versions:
    - images:
        - digest:
            message: >-
              Unable to retrieve digest for image:
              docker.io/kabanero/java-microprofile:0.2.25 associated with stack:
              java-microprofile 0.2.25. Error: MANIFEST_UNKNOWN: manifest
              unknown; map[Tag:v2.16]
          id: Eclipse MicroProfile®
          image: docker.io/kabanero/java-microprofile
...

```

Private images are handled with a secret in the same manner as GHE repos are handled (see #468 for requirements).
Sample secret:

```
kind: Secret
apiVersion: v1
metadata:
  name: mysecret
  namespace: kabanero
  annotations:
    kabanero.io/docker-0: 'https://docker.io'
data:
  password: <password | access token>
  username: <username>
```


deploy/crds/kabanero.io_stacks_crd.yaml
```diff
@@ -55,6 +55,15 @@ spec:
                     items:
                       description: Image defines a container image used by a stack
                       properties:
+                        digest:
```
kaczyns: I haven't looked at the rest of the code yet... but I think having the digest in the `spec` section is going to be confusing, and I'm guessing that we don't read it anywhere (or ignore it if it's there in the spec section).  I think that we need to think of a way to keep it out of the spec section.

pkg/controller/stack/stack_controller.go
```diff
@@ -576,7 +576,29 @@ func reconcileActiveVersions(stackResource *kabanerov1alpha2.Stack, c client.Cli
...
+				newStackVersionStatus.Images = append(newStackVersionStatus.Images, kabanerov1alpha2.ImageStatus{Id: img.Id, Image: img.Image})
+			}
+
+			// Retrieve stack image version activation digests. As such, it should only be done once during activation.
+			// If there is an error during first retrieval, a subsequent successful retry may set the current digest and
+			// not the activation digest. More precisely, the digest may not necessarily be the initial activation digest
+			// because we allow stack activation despite there being a failure when retrieving the digest and the
+			// image/digest may have changed before the next successful retry.
+			for j, image := range newStackVersionStatus.Images {
+				if len(image.Digest.Activation) == 0 {
```
kaczyns: Isn't the length of `image.Digest.Activation` always going to be zero, because you just copied `newStackVersionStatus.Images` from `curSpec.Images` and `curSpec.Images` doesn't have any digest information?  It seems like the logic here needs to also look at the current status to see if we already had a digest for that version of the stack (and image).

PR #585, opened 2020-03-26
targetNamespace

dacleyra:


pkg/controller/kabaneroplatform/targetnamespaces.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// List of Namespaces we want to bind
+	targetnamespaceList := k.Spec.TargetNamespaces
+	
+	// If targetNamespaces is empty, default to binding to kabanero
+	if len(targetnamespaceList) == 0 {
+		targetnamespaceList = append(targetnamespaceList, "kabanero")
+	}
+
+	/* Get all Rolebindings named kabanero-pipeline-deploy-rolebinding
```
kaczyns: This is correct - the controller runtime client doesn't want you looking at anything outside your namespace apparently.  I'm sure we're just using it wrong or something.  There are other places where we've used unstructured instead for this very reason.

pkg/controller/kabaneroplatform/targetnamespaces.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, targetnamespace := range targetnamespaceList {
+		// Check if the namespace exists
+		namespace := &unstructured.Unstructured{}
+		namespace.SetGroupVersionKind(schema.GroupVersionKind{
+				Group:   "",
+				Kind:    "Namespace",
+				Version: "v1",
+		})
+		err = cl.Get(ctx, client.ObjectKey{Namespace: targetnamespace, Name: targetnamespace,}, namespace)
+		if err == nil {
```
kaczyns: I guess I'm confused about this.... I would have thought that `cl.Get()` would return a NotFound error if it couldn't find it in the target namespace.  So you'd only want to create it if the NotFound was returned.

pkg/controller/kabaneroplatform/kabaneroplatform_controller.go
```diff
@@ -96,6 +97,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+	
+	// Index Rolebindings by name
+	if err := mgr.GetFieldIndexer().IndexField(&rbacv1.RoleBinding{}, "metadata.name", func(rawObj runtime.Object) []string {
```
kaczyns: I was curious if this is required, or if it's just a performance boost.  It's fine - I've just never used indexes before and was curious.

pkg/controller/kabaneroplatform/targetnamespaces.go
```diff
@@ -0,0 +1,130 @@
...
+					Kind:       k.TypeMeta.Kind,
+					Name:       k.ObjectMeta.Name,
+					UID:        k.ObjectMeta.UID,
+					Controller: &ownerIsController,
+				},
+			},
+		},
+		Subjects: []rbacv1.Subject{
+			rbacv1.Subject{
+				Kind: "User",
```
kaczyns: Should this be `Kind: "ServiceAccount"`?  I guess I don't really care, if it works this way it's fine.  I'm looking at the RBAC doc and it seems like there is a `ServiceAccount` kind.
https://kubernetes.io/docs/reference/access-authn-authz/rbac/

pkg/controller/kabaneroplatform/targetnamespaces.go
dacleyra: This is a check that the namespace exists before attempting to create the RoleBinding within the namespace. The check isn't entirely necessary. Can just attempt to create the RoleBinding and let it fail, and I had it that way at one point. I'm not even sold that doing Get is the right way to check, but it doesn't seem like the client package has opinionated methods for such things.

pkg/controller/kabaneroplatform/kabaneroplatform_controller.go
dacleyra: It was necessary when I was making the attempt with the structured List(). It would err about metadata.name not being indexed. I did not retry without when I realized I would have to use Unstructured.

pkg/controller/kabaneroplatform/targetnamespaces.go
dacleyra: yes, will fix

pkg/controller/kabaneroplatform/targetnamespaces.go
kaczyns: I'm sorry - I can't read :-) yes this is fine.

dacleyra:
Subject is corrected to Kind: ServiceAccount
Commented out Indexer, it is not necessary for Unstructured List, but was necessary if Stuctured, in case it ever changes back to Structured.

PR #586, opened 2020-03-26
Update the landing page deployment to use reencrypt route policy.

### LINKED ISSUE #178: Edge termination for landing page
Use edge TLS termination for the landing page instead of using certificates generated by the landing container. This way, on certain environments (e.g. IBM Cloud) the edge certificate is already trusted by the browsers.

mezarin:
Fixes #178

### DIFF
--- a/config/orchestrations/landing/0.1/kabanero-landing.yaml
+++ b/config/orchestrations/landing/0.1/kabanero-landing.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Service
 metadata:
   name: kabanero-landing
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: kabanero-landing-service-cert-secret
 spec:
   selector:
     app: kabanero-landing
@@ -19,7 +21,7 @@ spec:
     kind: Service
     name: kabanero-landing
   tls:
-    termination: passthrough
+    termination: reencrypt
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/pkg/controller/kabaneroplatform/landing.go
+++ b/pkg/controller/kabaneroplatform/landing.go
@@ -2,17 +2,19 @@ package kabaneroplatform
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/url"
 	"strings"
 
+	"github.com/go-logr/logr"
 	kabanerov1alpha2 "github.com/kabanero-io/kabanero-operator/pkg/apis/kabanero/v1alpha2"
 	"github.com/kabanero-io/kabanero-operator/pkg/controller/kabaneroplatform/utils"
 	kabTransforms "github.com/kabanero-io/kabanero-operator/pkg/controller/transforms"
-	"github.com/go-logr/logr"
-	mf "github.com/manifestival/manifestival"
+	"github.com/kabanero-io/kabanero-operator/pkg/versioning"
 	mfc "github.com/manifestival/controller-runtime-client"
+	mf "github.com/manifestival/manifestival"
 	consolev1 "github.com/openshift/api/console/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -70,6 +72,11 @@ func deployLandingPage(_ context.Context, k *kabanerov1alpha2.Kabanero, c client
 	}
 
 	err = m.Apply()
+	if err != nil {
+		return err
+	}
+
+	err = addTLSConfigToLandingRoute(k, c, rev)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


mezarin:
The fix for AdoptOpenJDK/openjdk-docker#256 fixed the issue with using the landing page as the init container. Having said that, for things to work, an update to the landing page's Liberty server.xml is required to pick up the new keystore. The Landing will be broken if either the kabanero or landing updates are not in at the same time.

@alohr51 @kaczyns how do you want to orchestrate shipping this?

After we decide how to orchestrate this, @alohr51 I have a commit with the landing change, do you want me to submit a PR with the update? Otherwise, this is the change that needs to be made:

Under: kabanero-landing/src/main/wlp/server.xml

Replace:
```
    <keyStore id="defaultKeyStore" password="changeit" />
    <ssl id="defaultSSLConfig" trustDefaultCerts="true" sslProtocol="TLSv1.2"/>
```
With This:
```
    <keyStore id="defaultKeyStore" 
              password="changeit" 
              location="/etc/tls/secrets/java.io/landingpage/keystores/keystore.p12"
              type="PKCS12"/>	
    <ssl id="defaultSSLConfig" keyStoreRef="defaultKeyStore" trustDefaultCerts="true" sslProtocol="TLSv1.2"/>
```

config/orchestrations/landing/0.1/kabanero-landing.yaml
```diff
@@ -19,7 +21,7 @@ spec:
     kind: Service
     name: kabanero-landing
   tls:
-    termination: passthrough
+    termination: reencrypt
```
kaczyns: Since this also requires changes to the landing page container, I think we need to make a `0.2` orchestration with this change.  Otherwise, if someone uses kabanero-operator 0.8.0 and codes `spec.version` to something like `0.7.0`, the landing page won't work.

pkg/controller/kabaneroplatform/landing.go
```diff
@@ -74,6 +76,11 @@ func deployLandingPage(_ context.Context, k *kabanerov1alpha2.Kabanero, c client
 		return err
 	}
 
+	err = addTLSConfigToLandingRoute(k, c, rev)
```
kaczyns: I think we'll need to see if we're using version `0.1` of the orchestration here, and if we are, skip this step.

config/orchestrations/landing/0.1/kabanero-landing.yaml
mezarin: Thanks Tim, Yeah that is very true. A bit more complicated issue is if someone deploys a CR instance with spec.version = 0.8.0 and decides to use landing version 0.7.0. In this case, the landing page won't come up, and it becomes a bit harder to seamlessly toggle the kabanero updates because we now need to introduce a restriction on the landing page version that can be used.

mezarin:
@kaczyns  Let me break down what needs to happen and where the problem area is. 

1. Generate TLS key/cert and put them in a secret (magic from annotation in the landing service)
2. Generate keystore from the key/cert (mounted to a pod )
3. Extract the TLS key key/cert from secret and add them to the landing route along with updating the tls termination policy.
4. Update the landing page's server.xml.

The problem here is step 3. To completely avoid the problem with mixed versions, at least, step 3 and 4 would have to be done by the landing page if it detected the presence of the generated security files.

config/orchestrations/landing/0.1/kabanero-landing.yaml
kaczyns: I understand - I consider that an edge case.  I think we need to support setting `spec.version` and expecting that the customer would take all of the rest of the defaults.  If we also support `spec.landing.version` then we can support the case where the customer has `spec.version` set to `0.8.0` but wants to use a prior version of the landing page for some reason.  The code that deploys the landing page would use the right version of the orchestration based on the value of `spec.landing.version` since the orchestration is coded into the version struct.  There would need to be special case code in `landing.go` that behaved differently based on the orchestration version.

kaczyns:
> @kaczyns Let me break down what needs to happen and where the problem area is.
> 
>     1. Generate TLS key/cert and put them in a secret (magic from annotation in the landing service)
> 
>     2. Generate keystore from the key/cert (mounted to a pod )
> 
>     3. Extract the TLS key key/cert from secret and add them to the landing route along with updating the tls termination policy.
> 
>     4. Update the landing page's server.xml.
> 
> 
> The problem here is step 3. To completely avoid the problem with mixed versions, at least, step 3 and 4 would have to be done by the landing page if it detected the presence of the generated security files.

I'm not sure I follow you here. Are you suggesting that the landing page edit the route that the operator is creating based on whether it detects that the certificates are present in the file system?

I had hoped we'd be able to support two cases:
1) landing page orchestration is 0.1.  We don't do anything with certificates.  the route gets created as it is today and the landing page server.xml is as it is today.  This would be version 0.7.0.
2) landing page orchestration is 0.2.  We generate service certificates using the openshift annotation.  We modify the route with those certificates.  The deployment spec we generate includes the init container which modifies the certificate files into a keystore (or we use the script in the liberty image that Dan suggested in Slack).  The landing page's server.xml expects the keystore to be created already and uses it.

Then there's case 3 where a user tries to use Kabanero v0.8.0 with some earlier version of the landing page, and I would expect that case to fail.  I think we could make that case work if they set `spec.version` to `0.8.0` and then just set `spec.landing.version` to `0.7.0` because we'd be able to detect that the landing page was using version `0.1` of the orchestration.  I'm pretty sure we can detect the orchestration version in the code.  We wouldn't be able to support cases like `spec.version` is set to `0.8.0` but then `spec.landing.image` is set to `kabanero/landing:0.7.0`..... but again I consider that an edge case, and if someone is setting specific images then they should know what they are doing and expect that they might need to work thru problems.

I may be missing something fundamental here... please let me know if I've done that.

mezarin:
The point with my two previous posts was that if we were going to support mixed versions, having a v1/v2 orchestrations will not cover all cases. Also, in order to cover all cases in a reliable fashion, the landing page, however feasible, would have to do more of the work as i mentioned in my second post.
Having said that, I think we agree v1/v2 orchestration will not cover all cases given your last post. 
Also what you call the edge case, you consider to be an acceptable failure case. I will go with that.

Last, I'll investigate if the script Dan mentioned will even help here given that it looks like it needs to be configured/enabled by the landing page's server.xml.

kaczyns:
We will come back and edit `version.yaml` to use the new orchestration when the landing page PR is merged.

PR #594, opened 2020-03-30
filtering pipelines archive/yaml to only tekton.dev resources

dacleyra:
Filter manifests from the pipeline archive to only Group: tekton.dev
Provide Info if a non tekton.dev manifest was found


dacleyra:
Yeah, I think so. Will try.

PR #604, opened 2020-04-03
pipelinesNamespace

dacleyra:
In-Progress

### DIFF
--- a/pkg/controller/kabaneroplatform/featured_stacks.go
+++ b/pkg/controller/kabaneroplatform/featured_stacks.go
@@ -68,6 +68,7 @@ func reconcileFeaturedStacks(ctx context.Context, k *kabanerov1alpha2.Kabanero, 
 					},
 					Spec: kabanerov1alpha2.StackSpec{
 						Name: key,
+						PipelinesNamespace: k.Spec.PipelinesNamespace,
 					},
 				}
 			} else {
--- a/config/samples/full.yaml
+++ b/config/samples/full.yaml
@@ -124,6 +124,9 @@ spec:
       sha256: e6ebc6aeaf21540f0d0dac8caf0a2d805e8d90f174cb912a31831f700d049bb1
       https:
         url: https://github.com/kabanero-io/kabanero-pipelines/releases/download/0.7.0/default-kabanero-pipelines.tar.gz
+        
+  # The namespace pipelines will be created in
+  pipelinesNamespace: kabanero-pipelines
 
   governancePolicy:
     # Provide governance configuration for all stacks managed by Kabanero. The allowed configuration policies are:
--- a/pkg/apis/kabanero/v1alpha2/kabanero_types.go
+++ b/pkg/apis/kabanero/v1alpha2/kabanero_types.go
@@ -46,6 +46,8 @@ type KabaneroSpec struct {
 	AdmissionControllerWebhook AdmissionControllerWebhookCustomizationSpec `json:"admissionControllerWebhook,omitempty"`
 
 	Sso SsoCustomizationSpec `json:"sso,omitempty"`
+	
+	PipelinesNamespace string `json:"pipelinesNamespace,omitempty"`
 }
 
 // InstanceStackConfig defines the customization entries for a set of stacks.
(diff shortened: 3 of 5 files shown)


dacleyra:
@kaczyns 

Can you take a look and see how this looks so far, thanks

---

Move the triggers Role out of the deploy script and into the orchestration, which gets applied along with the RoleBinding as it was
Move the stack-controller RBAC out of the deploy script and into the orchestration. The scope is changed to Namespace and not Cluster.
Allow the orchestration to create the namespace for pipelinesNamespace

Kabanero.Spec.PipelinesNamespace - the desired pipelinesNamespace
Kabanero.Status.PipelinesNamespace - set once the featured stacks have been resolved, and previous rbac is cleaned up
Stack.Spec.PipelinesNamespace - the desired pipelinesNamespace, filled during featured stack resolution
Stack.Status.PipelinesNamespace - set once the pipelines have been created

featured_stacks
if the K pipelinesNamespace has changed, and the stack has a different pipelinesNamespace, delete that Stack
At this step, the stack-controller still has the RBAC to act in the previous pipelinesNamespace

kabaneroplatform_controller
reconcilePipelinesNamespace cleans up the RBAC for the old pipelinesNamespace once there are no more stacks present in that namespace

stack-controller
Create/Delete the RBAC necessary for supporting stack-controller & pipelineNamespace

stack_controller
create pipelines in the pipelinesNamespace instead of the stack namespace


TODO
- cleanup some debug logs
- disable rbac creation in deploy script that is now present in the orchestration

pkg/controller/kabaneroplatform/stack-controller.go
```diff
@@ -219,3 +298,75 @@ func getStackControllerStatus(ctx context.Context, k *kabanerov1alpha2.Kabanero,
...
+	deployedStacks := &kabanerov1alpha2.StackList{}
+	err := c.List(ctx, deployedStacks, client.InNamespace(k.GetNamespace()))
+	if err != nil {
+		return err
+	}
+
+	// Remove the old SA/Role/Bindings once all the Featured Stacks are re-created in the new pipelinesNamespace
+	canCleanup := true
+	for _, deployedStack := range deployedStacks.Items {
+		if k.Spec.PipelinesNamespace != deployedStack.Status.PipelinesNamespace {
```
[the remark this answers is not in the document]
kaczyns: this was my mistake, I was using the wrong version of the admission webhook and stack controller, which were removing the `pipelinesNamespace` from the stack object and therefore causing the stacks to be deleted.

pkg/controller/kabaneroplatform/stack-controller.go
```diff
@@ -219,3 +298,75 @@ func getStackControllerStatus(ctx context.Context, k *kabanerov1alpha2.Kabanero,
...
+			canCleanup = false
+		}
+	}
+	
+	if canCleanup == true {
```
kaczyns, line 335: I see what you are doing here.... I had not considered this before.  Yes we need to leave permissions around long enough for the stack controller to clean up the pipelines, etc, in the old namespace.  However when I tried switching the namespace on my cluster, this did not behave as I expected.
Old pipelinesNamespace was `fred`, new was `kabanero`.  First time thru, it created stuff in `kabanero` and I could see things transitioning to `kabanero`.  But then the kabanero controller reconcile was driven again, and this time it deleted all of the roles and bindings from `kabanero`.  I think this code needs to be able to detect that there is cleanup work to do for a namespace, and then when it's done, put a mark on the wall or something so that it doesn't try to do it again.

pkg/controller/kabaneroplatform/featured_stacks.go
```diff
@@ -168,8 +175,26 @@ func preProcessCurrentStacks(ctx context.Context, k *kabanerov1alpha2.Kabanero,
 		return err
 	}
 
+	// Only keep the FeaturedStack if the Kabanero pipelinesNamespace did not change, otherwise delete & recreate
+	pipelinesNamespace := pipelinesNamespace(k)
+	for _, deployedStack := range deployedStacks.Items {
+		if deployedStack.Spec.PipelinesNamespace != pipelinesNamespace {
+			err := cl.Delete(ctx, &deployedStack)
```
kaczyns, line 182: Before we delete a stack with the incorrect pipeline namespace, we have to be sure that none of the versions have `desiredState` set to some value.  The `desiredState` is a clue that the stack is being manually managed.  I believe there is other code in here that checks the `desiredState` when attempting to sync up with the stack hub index yaml.

PR #621, opened 2020-04-13
Prevent activation digest from being cleared.

mezarin:
Updates to already deployed stacks may cause the reported activation digest (status.versions[x].images[x].digest.activation) to be cleared:

...
[{Eclipse MicroProfile® docker.io/kabanerobeta/java-microprofile {37cdf72ab9589e4b0f1389e1c6acc574a1aa8c9cc2234f6be3fd2e994e1de93f }}]

...
 [{Eclipse MicroProfile® docker.io/kabanerobeta/java-microprofile { }}]


mezarin:
This fix should be ported to 0.8.0-rc.*

PR #636, opened 2020-04-17
Image URL parser needs to handle a digest in addition to a tag

dacleyra:
WIP

I'm not familiar with why we were doing the parsing heavy lifting in utils/image.go when there is the reference parser.

However, there is a behavioral change, but I'm not sure if the downstream impact is of concern, but they are picked up by the existing test case.

1. ParseNormalizedName will give us a docker.io prefix when no domain is specified.
2. ParseNormalizedName will not replace a malformed domain with docker.io

https://godoc.org/github.com/docker/distribution/reference

Let me know what you think


kaczyns:
Right - it was a conscious choice to not use the reference parser.  The first reason was just that it wasn't clear to us at the time if it was a good idea to use the Docker reference parser against other types of image registries - I think it's pretty clear now that it's OK.  But the second was some of the behavior that the tests uncover when using the reference parser.  Sometimes the reference parser pre-pends with the default hostname.  At the time there was string matching going on by the Kab Pipelines in verifying whether the desired build was for an active stack.  Running the image name that was in the stack, thru the parser to strip the tag off, was returning a different registry/repository than what was put in, which complicated the matching code that the Pipelines component had to do.  Since our use cases at the time were to:
1) strip off the tag
2) parse the hostname for secret matching
it seemed like just doing it ourselves was going to give us the best shot at keeping the behavior sensible (stripping off the tag should not modify the hostname).  

Now I'm not opposed at this point to using the reference parser - but if we do it, then there is going to need to be a lot of coordination and testing.  We'll need to get with the kab pipelines team, and the eventing team, to make sure that they will be able to tolerate us inserting the hostname (sometimes) in the Stack CR instances status.  And also to make sure that our secret-scanning code still work (that we'll be able to find the correct secret when logging into docker hub).  I think the latter is less likely to be a problem because the secret matching code requires us to have the secret annotated with a hostname.

dacleyra:
The stack index, and .appsody-config.yaml all have it formatted as ex: 

docker.io/kabanero/java-openliberty:tag

The test pipeline runs execute correctly

So, not breaking anything I'm aware of. Maybe if no hostname is used? That said, there is no kabanero-stack-hub release that has been published without the hostname set.

The test cases for undefined hostname still need to be modified

@smcclem any concerns?

dacleyra:
Updated tests. This should make it clear where the parsing behavior ends up with differences.

PR #649, opened 2020-04-24
Add ingress CA cert to http transport root CAs when making http calls.

### LINKED ISSUE #638: Need to trust default ingress certificate
In addition to being able to trust certificates signed by well-known CAs (such as those presented by sites like github.com), the operator will need to be able to trust the OpenShift default ingress certificate.  When the stack hub is hosted locally, it is likely that the certificate it presents to clients is the default ingress certificate, and if the operator trusts this, it does not need to disable certificate verification checking for all hosts.

We may also wish to consider trusting the CA that generates service serving certificates (if we don't trust this already).

mezarin:
Fixes #638 
Preliminary test and test using ICPA completed successfully.


pkg/controller/stack/stack_controller_test.go
```diff
@@ -1787,8 +1780,8 @@ func TestReconcileActiveVersionsSkipCertVerify(t *testing.T) {
 		t.Fatal(fmt.Sprintf("Should be an error in the status message"))
 	}
 
-	if !strings.Contains(stackResource.Status.Versions[0].StatusMessage, "x509") {
```
kaczyns, line 1790: This is really the only issue that I have - we need a way to be able to test that we do or do not trust the certificate that the server presents, and that's what this test was doing.  Now it just sees that `router-ca` is not present.  

We need some way to proceed when `router-ca` is not present, and the client asked us to verify the server's certificate.  Some ideas:
1) If `router-ca` is not there, issue an error/warning message to the log and proceed without it.
2) Capture a `router-ca` and have the controller-runtime client for this test case return it when the code asks for it.

If we're going to insist on `router-ca` being there, then we should probably test what happens when someone replaces the default ingress certificate with a custom one (making sure `router-ca` now has the custom certificate). See:
https://docs.openshift.com/container-platform/4.3/authentication/certificates/replacing-default-ingress-certificate.html

kaczyns:
I resolved merge conflicts, and added a commit that makes `router-ca` optional.  I have not tested this yet.

kaczyns:
I guess I should explain why I did this.... I did some searching about `router-ca` and it seems like it's meant to be an internal (not an external).  Also it seems like `router-ca` is not necessarily updated, if a customer was to change to use a custom certificate, and no ingresses were using the default certificate.  I came to that conclusion after reading this bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1772775

I think that at this point making it optional is safer.  Some user may ask us to support adding the custom certificate to the trust chain, and when that requirement comes in we'll have to learn how to read the default ingress and see if a custom certificate was provided, and then read that certificate instead.  Or perhaps just let the user add a list of certificates we should trust to the CR somehow.

pkg/controller/utils/cache/http.go
```diff
@@ -40,26 +42,44 @@ func GetHTTPClient(accessToken []byte, transport *http.Transport) (*http.Client,
...
 
-// Populates a TLS config struct based specified input.
-func GetTLSCConfig(c client.Client, skipCertVerify bool) (*tls.Config, error) {
+// Log mutex
+var logRouterCertError sync.Once
+
+// Log that there was a problem obtaining the default ingress router CA
+// certificate.  Only log once as the same error is likely to happen
+// over and over again.
+func logIngressRouterCertError(logger logr.Logger, err error) {
```
mezarin, line 51: Tim, i am concerned that printing it once may get lost and it would become hard to debug. If possible, could we leave a breadcrumb of information in the error message when we detect that request failed? That is by labeling/printing the client/transport struct in the error message.

kaczyns: @mezarin I just want to be sure I understand what you are asking for.... you want to print the golang struct for the client and transport in the error message that we print to the pod log?

I wasn't sure if you were referring to the pod log or the CR log.

mezarin: Hi Tim, It would be the kabanero stack controller pod log. If there is a failure with the http request that used transport.TLSClientConfig that is nil or not ... the error we print when the request failed should include it so we know if it was set or not.

kaczyns: Hi Ed - it was pretty easy to make this change to the HTTP path.  I did not make the change to the GitHub path because the TLSClientConfig check is a couple function calls beneath where the client is used.  We can make that change later if we decide it's necessary - maybe we can wrap the GitClient or something so that we can retrieve the TLS config later.

Issue #536, opened 2020-02-27
Kabanero operator should set up Codewind tekton roles


johnmcollier:
(Originally https://github.com/eclipse/codewind/issues/2333)

In order to properly launch the tekton dashboard from the "Open Tekton Dashboard" button in Codewind/Theia, Codewind needs to be set up with a cluster role and rolebinding allow it to retrieve the tekton dashboard service in the tekton-pipelines namespace.

You can see the files here:
- https://github.com/eclipse/codewind-che-plugin/blob/master/setup/install_che/codewind-tektonrole.yaml
- https://github.com/eclipse/codewind-che-plugin/blob/master/setup/install_che/codewind-tektonbinding.yaml

This issue isn't a stopship and so my fix (https://github.com/kabanero-io/kabanero-operator/pull/535) should only be merged after release.

mtamboli:
@kaczyns where would this defect be fixed: install script or kabanero operator?

jgawor:
@johnmcollier is there a work-around that doesn't require operator updates?

johnmcollier:
@jgawor The following command could be used as a workaround until the operator is updated:
```
kubectl apply -f https://gist.githubusercontent.com/johnmcollier/41031692b8bfe067ef28223069e57871/raw/00cf207e1f9f3b6046de61b278b7b5f96f1a92b0/kabanero-codewind-tekton.yaml
```

**Note:** We would want to move the yaml off of a github gist and somewhere more permanent.

marikaj123:
According to Jarek, this issue does not require to be fixed in 0.6.0.

marikaj123:
@johnmcollier - Does Codewind/CodeReady workspace have the fix? If yes, what would be the version that would contain the fix?

johnmcollier:
@marikaj123 Not yet, it's a fix I need to make in the Kabanero operator. I can deliver it for Kabanero 0.9.0

marikaj123:
@johnmcollier - Thank you for getting it fixed for Kabanero 0.9.0.  The fix will need to be complete during the month of April to make it into Kabanero 0.9.0.

marikaj123:
@johnmcollier -  Are you producing the che-devfile-registry image that in turn pulls in the correct version of codewind to allow the operator to include the codewind 0.11.0?

PR #653, opened 2020-04-27
Add roles for Tekton and add updates for Codewind 0.11.0/CRW 2.1

johnmcollier:
This PR tackles a few items:
- Adds a role and rolebinding in the `tekton-pipelines` namespace, allowing the tekton dashboard to be accessed from CRW (fixes https://github.com/kabanero-io/kabanero-operator/issues/536)
- Renames the cluster role used for CRW, to prevent an issue when Kabanero and Eclipse Che are installed on the same cluster (and use the same cluster role for codewind)
- Updates the role for Codewind, allowing it to start in per-user namespaces in CRW, when openshiftOAuth is enabled

### DIFF
--- a/pkg/controller/kabaneroplatform/codereadyworkspaces.go
+++ b/pkg/controller/kabaneroplatform/codereadyworkspaces.go
@@ -484,6 +484,16 @@ func deleteCRWOperatorResources(ctx context.Context, k *kabanerov1alpha2.Kabaner
 		if err != nil {
 			return err
 		}
+
+		// Delete the Tekton role and rolebinding too
+		err = processCRWYaml(ctx, k, rev, unstructured.Unstructured{}.Object, c, crwYamlNameCodewindTektonRole, false, "tekton-pipelines")
+		if err != nil {
+			return err
+		}
+		err = processCRWYaml(ctx, k, rev, unstructured.Unstructured{}.Object, c, crwYamlNameCodewindTektonBinding, false, "tekton-pipelines")
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/controller/kabaneroplatform/codereadyworkspaces.go
```diff
@@ -67,13 +69,27 @@ func reconcileCRW(ctx context.Context, k *kabanerov1alpha2.Kabanero, c client.Cl
...
-	err = processCRWYaml(ctx, k, rev, templateCtx, c, crwYamlNameCodewindClusterRole, true)
+	// Deploy the Codewind cluster role with the required permissions for codewind.
+	err = processCRWYaml(ctx, k, rev, templateCtx, c, crwYamlNameCodewindClusterRole, true, k.GetNamespace())
 	if err != nil {
 		logger.Error(err, fmt.Sprintf("Failed to Apply clusterRole resource. Revision: %v. TemplateCtx: %v", rev, templateCtx))
 		return err
 	}
 
+	// Deploy the Codewind Tekton role
+	err = processCRWYaml(ctx, k, rev, templateCtx, c, crwYamlNameCodewindTektonRole, true, "tekton-pipelines")
```
kaczyns, line 80: The things that are created in a namespace (in this case the `tekton-pipelines` namespace) can't have an owner reference back to an object in another namespace (in this case the Kabanero CR instance in the `kabanero` namespace).  This means that we'll need cleanup logic to delete the role when either Codewind is disbaled, or when the Kabanero CR instance is deleted.  I think the `cleanupCRW` function does this already.  It calls `deleteCRWOperatorResources()` which deletes the cluster role.  So I think you'd just have to additionally delete this one (and the one on line 87 which is also in `tekton-pipelines`.

johnmcollier: @kaczyns Thanks Tim, I'll take a look tomorrow.

mezarin:
Looks ok. Merging.

PR #662, opened 2020-04-29
Add option to use Docker images from master branch

hbtran:


kaczyns:
Hi @hbtran - thanks for submitting this PR.  I understand that special processing is needed to work from the master branch, but I sort of object to having branch-specific code in the shell scripts.  The scripts already let the user define environment variables to control URL of the subscriptions and custom resource yamls.  I don't understand why your automation / scripting can't `curl` the subscriptions yaml, do the `sed`, and then run the script passing in two environment variables that point to the modified subscription yaml and the custom resource yaml?

I agree that we'd need a way to let the user influence what we print for the example Kabanero CR instance yaml (default.yaml).

The two common use cases we're trying to support are:
1) User installing from a tagged release (in GitHub)
2) User installing from code they've pulled down to their workstation.

Installing directly from master via the shell script is neither of these, and is something that we'd be unable to test on a regular basis to determine if we changed something that would break the ability to install directly from master.

hbtran:
I had tested an alternative method.  Could we add another CatalogSource to the kabanero-subscriptions.yaml, labeled 00-catalogsource-master pointing to the Docker master image?  That way, we can add an optional environment variable to override the CatalogSource install selector.  There won't be any special processing anywhere and also @garypicher  wanted users to be able to install from master.   It might interfere with the Makefile though.

kaczyns:
@mezarin is working on another solution that is more like your first solution.  I'd like to keep the install script as it is, and we're going to have a second script that is never attached to a release, which is intended to do an install from master.

The workflow would be:
1) clone kabanero-operator github repo
2) run the new shell script (no need to set any environment variables)

This will install kabanero using whatever is tagged `master` out in docker hub.  

I will leave this PR open until we've agreed that his solution meets your requirements.

kaczyns:
@hbtran We just merged a `deploy/install-master.sh` script.  Would you please try it out and let us know if it suits your needs?

hbtran:
Yes, it's works, thank you!

kaczyns:
Awesome - I will close this PR.

PR #667, opened 2020-04-30
Move tekton dashboard back to tekton-pipelines namespace

### LINKED ISSUE #665: Move tekton dashboard 0.6.1 back to `tekton-pipelines` namespace
The Tekton dashboard has moved to `openshift-pipelines`.  There is Kabanero runtime code that assumes it is in the `tekton-pipelines` namespace.  For the time being, move the dashboard install back to `tekton-pipelines`.

kaczyns:
Fixes #665 
Version 0.6.1 of the Tekton dashboard moved (for OpenShift) from `tekton-pipelines` to `openshift-pipelines` namespace.  Since the Kabanero runtime expects the dashboard to be in the `tekton-pipelines` namespace, for the time being we are going to move it back there.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -318,8 +318,8 @@ oc new-project tekton-pipelines || true
 
 openshift_master_default_subdomain=$(oc get ingresses.config.openshift.io cluster --output=jsonpath={.spec.domain})
 
-curl -s -L https://github.com/tektoncd/dashboard/releases/download/v0.6.1/openshift-tekton-webhooks-extension-release.yaml | sed "s/{openshift_master_default_subdomain}/${openshift_master_default_subdomain}/" | oc apply -f -
-oc apply -f https://github.com/tektoncd/dashboard/releases/download/v0.6.1/openshift-tekton-dashboard-release.yaml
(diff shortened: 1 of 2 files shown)


deploy/install.sh
```diff
@@ -318,8 +318,8 @@ oc new-project tekton-pipelines || true
 
 openshift_master_default_subdomain=$(oc get ingresses.config.openshift.io cluster --output=jsonpath={.spec.domain})
 
-curl -s -L https://github.com/tektoncd/dashboard/releases/download/v0.6.1/openshift-tekton-webhooks-extension-release.yaml | sed "s/{openshift_master_default_subdomain}/${openshift_master_default_subdomain}/" | oc apply -f -
-oc apply -f https://github.com/tektoncd/dashboard/releases/download/v0.6.1/openshift-tekton-dashboard-release.yaml
+curl -s -L https://github.com/tektoncd/dashboard/releases/download/v0.6.1/openshift-tekton-webhooks-extension-release.yaml | sed "s/{openshift_master_default_subdomain}/${openshift_master_default_subdomain}/" | sed "s/openshift-pipelines/tekton-pipelines/" | oc apply -f -
```
mezarin, line 321: Hi Tim, I think the sed is a bit too generic. You do not want to update other entries with openshift-pipelines in the value entries (i.e. see: WEBHOOK_CALLBACK_URL). If you do, ok. Otherwise, making the sed more specific to namespace would be better.

kaczyns: Actually that URL contains the namespace in it, that's why it needs to change too.

mezarin: Ok thanks. Will approve.

PR #671, opened 2020-05-01
update console's kabaneros permissions

### LINKED ISSUE #670: Console needs elevated permissions to update CRD
When we try to update the Kabanero CRD using the operator bindings library we get back a `Forbidden` message and the CR is not updated. We need enhanced permissions (I think [here](https://github.com/kabanero-io/kabanero-operator/blob/master/config/orchestrations/landing/0.1/kabanero-landing.yaml#L42) to allow us to create, update, and delete the CR if needed.

The Console may in the future need to read configMaps as well, not sure if that needs to be an added permission as well.

More info

You can view the code we use to update the instance using the operator bindings library [here](https://github.com/kabanero-io/kabanero-landing/blob/2a1ecffbada163f5322eeaab3cf4a1358622ef7e/src/main/java/io/kubernetes/KabaneroClient.java#L133-L143). The catch block executes with the message
>[WARNING ] Error with setting kabanero instance kabanero
Forbidden

When running locally we didn't have a problem updating the CRD on the cluster, but when the app is running on OpenShift this becomes a problem.

alohr51:
update console's kabaneros permissions to allow create,delete,update

closes #670

### DIFF
--- a/config/orchestrations/landing/0.1/kabanero-landing.yaml
+++ b/config/orchestrations/landing/0.1/kabanero-landing.yaml
@@ -74,4 +74,4 @@ rules:
   - get
   - list
   - watch
-  
+
--- a/config/orchestrations/landing/0.1/kabanero-landing.yaml
+++ b/config/orchestrations/landing/0.1/kabanero-landing.yaml
@@ -74,7 +74,3 @@ rules:
   - get
   - list
   - watch
-  - create
-  - update
-  - delete
-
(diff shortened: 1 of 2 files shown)


alohr51:
I just realized there are 2 orchestrations so I'll have to add a commit to update both

alohr51:
should we leave the 0.1 orchestration alone?

PR #672, opened 2020-05-01
kab landing to 0.9.0-rc.1

alohr51:


alohr51:
I realize there are 2 orchestrations so I'll have to add a commit to update both

PR #682, opened 2020-05-05
Update stack finalizer logic

### LINKED ISSUE #659: Stack finalizer does not complete when unknown tekton types in pipelines
If I have a pipeline zip, with an invalid tekton object type in it, then I get the following in my Stack CR status:
```yaml
      - assetDigest: fc451d39a3dacac4e82044d1195a5553115e5953965326bc41227c0d1bba4d6a
        assetName: java-microprofile-build-push-pl-template
        group: tekton.dev
        kind: TriggerTemplate
        namespace: tekton-pipelines
        status: unknown
        statusMessage: 'Unable to check asset: no matches for kind "TriggerTemplate"
          in version "tekton.dev/v1alpha1"'
        version: v1alpha1
```
This is a good error message.  However when I try to delete the stack, or the Kabanero CR instance, the stack controller finalizer never runs.  I suspect it's because it's trying to check and see if it needs to create this object, and it can't, because the type does not exist.  So when it checks, it gets an error every time.

Some solutions I can think of:
1) see if we can detect the "no kind" error and just assume it's not there (because the kind does not exist)
2) skip assets with `status` not equal to `active`  (need to be good about making sure the `status` is correct when the assets are activated)  (maybe can do this conditionally if the delete fails)

This is particularly a problem now, because Tekton triggers 0.4 changed the api group, and so if you try to activate old pipelines, you fall into this case (because now `TriggerTemplate` is not found in the old `apiGroup`).

mezarin:
Fixes #659

### DIFF
--- a/pkg/controller/kabaneroplatform/stack-controller.go
+++ b/pkg/controller/kabaneroplatform/stack-controller.go
@@ -150,7 +150,7 @@ func cleanupStackController(ctx context.Context, k *kabanerov1alpha2.Kabanero, c
 	}
 
 	templateCtx := rev.Identifiers
-	templateCtx["name"] = "kabanero-" + k.GetNamespace() + "-trigger-rolebinding"
+	templateCtx["name"] = "kabanero-" + k.GetNamespace() + "-stack-trigger-rolebinding"
 	templateCtx["kabaneroNamespace"] = k.GetNamespace()
 
 	f, err := rev.OpenOrchestration("stack-controller-tekton.yaml")
--- a/pkg/controller/kabaneroplatform/stack-controller.go
+++ b/pkg/controller/kabaneroplatform/stack-controller.go
@@ -5,10 +5,10 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/kabaneroplatform/stack-controller.go
```diff
@@ -78,7 +78,7 @@ func reconcileStackController(ctx context.Context, k *kabanerov1alpha2.Kabanero,
 	// Create a RoleBinding in the tekton-pipelines namespace that will allow
 	// the stack controller to create triggerbinding and triggertemplate
 	// objects in the tekton-pipelines namespace.
-	templateCtx["name"] = "kabanero-" + k.GetNamespace() + "-trigger-rolebinding"
+	templateCtx["name"] = "kabanero-" + k.GetNamespace() + "-stack-trigger-rolebinding"
```
kaczyns, line 81: I think a similar change is needed in `cleanupStackController()`.

mezarin: Hi Tim, yep it most definitely needs the update. Thanks for catching it. I made the update.
Need to go see what i did to my finder tool that did not find it when i searched the code.

PR #690, opened 2020-05-05
serverless GA, openshift-pipelines-operator.v0.11.2

dacleyra:
Change serverless to install from redhat catalog, as it is now GA
update to openshift-pipelines-operator.v0.11.2, still canary


deploy/kabanero-subscriptions.yaml
```diff
@@ -75,10 +75,10 @@ metadata:
   labels:
     kabanero.io/install: 12-subscription
 spec:
-  channel: kabanero-0.8
+  channel: "4.4"
```
kaczyns, line 78: I noticed in the OpenShift doc that for OCP 4.3 they want us to use the "4.3" channel but for OCP 4.4 they want us to use the "4.4" channel.  I'm guessing that at some point they may make updates to the 4.4 channel that are not compatible with OCP 4.3.  We should think about whether there is a way we can pick the correct channel based on the OCP version.  We'd also have to detect an upgrade from 4.3 to 4.4 and do the right thing.  This is fine for now.

Issue #702, opened 2020-05-11
kabanero operator reconcile problem if namespace is created after rolebinding


tam512:
Use Kabanero 0.9.0-rc.2, run pipeline nodejs-express-build-deploy-pl in Tekton dashboard to deploy an application to svt-project2 namespace.   I added svt-project2 to `spec.targetNamespaces` in Kabanero CR, but I forgot to create svt-project2 the namespace so the deploy-image step failed.
So I created the svt-project2 namespace and reran the pipeline, but it still failed because the namespace was created afterwards, so it didn't loop back around to create the rolebinding for permission again.

The current workaround is to restart kabanero operator pod (_oc -n kabanero delete pod -l name=kabanero-operator_)

Below is the error on deploy-image step
```
Error from server (Forbidden): error when retrieving current configuration of:
Resource: "appsody.dev/v1beta1, Resource=appsodyapplications", GroupVersionKind: "appsody.dev/v1beta1, Kind=AppsodyApplication"
Name: "radcl-njsexpress-app", Namespace: "svt-project2"
[...]

PR #717, opened 2020-05-14
Retry targetnamespace processing if failure

kaczyns:
This fixes part of #702.  There is a CLI service role binding that I'd also like to add here later.
This should not get reviewed/merged until 0.10.0 (after release-0.9 branch is created).

If a user adds a `targetNamespace` to the Kabanero CR instance, but the namespace does not exist yet, we do not come back and try to create it again.  I added a failure to the reconcile for this case so we'll come back and try again.

I also changed the way we process the namespaces.  The existing logic would not work if we had multiple Kabanero instances on the cluster, because we were assuming that any role binding with a given name, in any namespace, belonged to us.  Now, we're keeping track of the namespaces we've modified in the `Status`, and only replace them when we've determined that the new set of target namespaces is valid (the namespaces exist).


kaczyns:
@dacleyra would you mind taking another look at this please?  I think the most recent commit addresses your comments.  We are watching namespace objects now (only create and delete, not update).  When we detect a namespace create or delete we'll look to see if we have any Kabanero instances that want to use that namespace, and we'll reconcile those instances.

I've also changed the logic in targetnamespaces.go to allow us to apply a target namespace even if other target namespaces don't exist.

PR #727, opened 2020-05-20
Remove v1alpha1 CRDs and related bits

dacleyra:


kaczyns:
This looks good Dan.  I didn't see anything that would delete the collection controller deployment and friends on an upgrade from 0.9.0 to 0.10.0.  I think we need something in the kabanero controller that makes sure the collections stuff is deleted (deployment/pod, SA, roles, etc).  I'm not sure that stuff would get deleted on an upgrade unless the Kab CR instance was deleted.  Do you agree?
@dacleyra

dacleyra:
Maybe. The upgrade scenario doesn't satisfy InstallPlan

```
    Message:               error missing existing CRD version(s) in new CRD: kabaneros.kabanero.io: New CRD (kabaneros.kabanero.io) must contain existing served versions
(v1alpha1)
```

https://github.com/operator-framework/operator-lifecycle-manager/blob/fb2879e6ec1f54973f69a2d6136fc99094e2b261/doc/design/dependency-resolution.md#deprecateremove-a-version-of-crd

I don't think we set v1alpha1 served: false in a version previous

kaczyns:
Do we just need to introduce an artificial step (like have 0.9.1 set served = false for v1alpha1 and then 0.10.0 can remove it)?

dacleyra:
yes

dacleyra:
Setting served = false in 0.9.1, and then installing 0.10.0 from the PR worked
InstallPlan was successful
collection controller deployment is gone
the collections CRD sticks around though, but that seems to be normal? I don't think OLM necessarily cleans out CRDs

dacleyra:
close in favor of 
https://github.com/kabanero-io/kabanero-operator/pull/760

Issue #734, opened 2020-05-27
Updates to kabanero-pipeline-role for signing


mtamboli:
We need to add more permissions to  `kabanero-pipeline-role` so that push command works after container signing is enabled. @toshiyamamoto please provide exact details.

mtamboli:
@marikaj123 we need this for 0.9.1 if possible.

toshiyamamoto:
In order to put and get signatures in the internal image registry and external signature store, the following role should be granted to the service account:
- apiGroups:
  - ''
  - image.openshift.io
  resources:
    - imagesignatures
  verbs:
    - create
    - delete
- nonResourceURLs:
  - '/staging/*'
  verbs:
    - get
    - put

mtamboli:
Verified the fix. Closing.

PR #735, opened 2020-05-27
add imagesignature role

toshiyamamoto:
To support pulling images which are signed from the internal image registry, imagesignature role needs to be granted to kabanero-pipeline-role.

### DIFF
--- a/deploy/kabanero-customresources.yaml
+++ b/deploy/kabanero-customresources.yaml
@@ -262,6 +262,7 @@ rules:
   verbs:
     - create
     - delete
+    - get
 - nonResourceURLs:
   - '/staging/*'
   verbs:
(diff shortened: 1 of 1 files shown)


toshiyamamoto:
fixes #734

deploy/kabanero-customresources.yaml
```diff
@@ -254,6 +254,20 @@ rules:
...
+  - ''
+  - image.openshift.io
+  resources:
+    - imagesignatures
+  verbs:
+    - create
+    - delete
+    - get
+- nonResourceURLs:
+  - '/staging/*'
```
jgawor, line 267: @toshiyamamoto can you explain the `/staging/` permission?

toshiyamamoto: When an external image registry with sample look aside signature server is used, after signing an image, a sign script puts a generated signature to a signature server by using curl, and k8s checks whether outgoing http request is allowed by checking the service account roles.  This specific uri is used for accessing the sample signature server. 
If the customer has their own signature server and it's url is different from the sample, then they need to update the role to match their environment.

PR #741, opened 2020-05-29
Add GrafanaDashboard, Secret and ConfigMap to pipeline ClusterRole, fix others

ianpartridge:
cc: @kvijai82

### DIFF
--- a/deploy/kabanero-customresources.yaml
+++ b/deploy/kabanero-customresources.yaml
@@ -367,7 +367,7 @@ rules:
 - apiGroups:
   - integreatly.org
   resources:
-  - grafanaDashboard
+  - grafanadashboards
   verbs:
   - get
   - list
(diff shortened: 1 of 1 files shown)


ianpartridge:
@kvijai82 why do some of the `resources` in this ClusterRole specify the plural name (`appsodyapplications`, `serviceBindingRequests`) and some specify the singular name (`serviceMonitor`, `database`)?  Does it matter?

kvijai82:
@ianpartridge that's the change I am trying to test out with Terrance today and will be creating a PR for.  It does matter.

ianpartridge:
If it matters, do they have to be singular or plural?  I would guess plural, because that's what we had for the existing resources (`appsodyapplications`, `openlibertyapplications`).  If I'm correct, I guess that means I should update this PR to say grafanaDashboards?

ianpartridge:
I saw that the CR you sent @tnixa was plural, so I've updated this PR accordingly. Hope that's correct.

ianpartridge:
Also added `Secret` and `ConfigMap` as two core resources we expect users will want to include in their GitOps repo. `Secret` is definitely required in order to get OpenLiberty metrics working (because the `/metrics` endpoint requires authentication).

